PR #1, opened 2025-10-28
fix: stabilize property setter functions by using global options & fix ci

mfazekas:
## Summary

This PR fixes the issue where property setter functions from `useRive*` hooks were unstable, causing unnecessary re-renders.

### Changes
- Extract options objects to global constants in `useRiveBoolean`, `useRiveNumber`, `useRiveString`, and `useRiveColor`
- Use `useMemo` in `useRiveTrigger` to recreate options only when `onTrigger` changes
- Remove unnecessary `useCallback` imports where no longer needed
- fixed GitHub actions  (yarn nitrogen was missing from few places)

### Problem
Property setter functions (setValue, setButtonText, etc.) were unstable because the options object passed to `useRiveProperty` was recreated on each render, even though `useCallback` was used for the `getProperty` function. This caused the `useMemo` and `useEffect` in `useRiveProperty` to re-run unnecessarily, resetting listeners and creating instability.

### Solution
By extracting the options to global constants, we ensure stable object references across renders. For `useRiveTrigger`, which has a dynamic `onTrigger` parameter, we use `useMemo` to only recreate the options when `onTrigger` actually changes.

## Test plan
To reproduce the issue: In [example/src/pages/RiveDataBindingExample.tsx](https://github.com/rive-app/rive-nitro-react-native/blob/main/example/src/pages/RiveDataBindingExample.tsx), apply this diff:

```diff
diff --git a/example/src/pages/RiveDataBindingExample.tsx b/example/src/pages/RiveDataBindingExample.tsx
index 76b2f60..039236c 100644
--- a/example/src/pages/RiveDataBindingExample.tsx
+++ b/example/src/pages/RiveDataBindingExample.tsx
@@ -1,4 +1,4 @@
-import { View, Text, StyleSheet, ActivityIndicator } from 'react-native';
+import { View, Text, StyleSheet, ActivityIndicator, Button } from 'react-native';
 import { useState, useEffect } from 'react';
 import {
   Fit,
@@ -152,6 +152,10 @@ export default function DataBindingExample() {
             hybridRef={setHybridRef}
           />
         )}
+        <Button
+          title="Change button"
+          onPress={() => setButtonText('New Button Text')}
+        />
       </View>
     </View>
   );
```
pressing the change button should change the button for good. Previously it was changed, but changed back immediately

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,9 @@ jobs:
       - name: Lint files
         run: yarn lint
 
+      - name: Codegen
+        run: yarn nitrogen
+
       - name: Typecheck files
         run: yarn typecheck
 
@@ -46,6 +49,9 @@ jobs:
 
       - name: Setup
         uses: ./.github/actions/setup
+
+      - name: Build package
+        run: yarn nitrogen
 
       - name: Build package
         run: yarn prepare
@@ -100,6 +106,10 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
 
+      - name: Generate nitrogen code
+        if: env.turbo_cache_hit != 1
+        run: yarn nitrogen
+
       - name: Build example for Android
         env:
           JAVA_OPTS: "-XX:MaxHeapSize=6g"
@@ -109,7 +119,7 @@ jobs:
   build-ios:
     runs-on: macos-latest
     env:
-      XCODE_VERSION: 16.2
+      XCODE_VERSION: 16.4
       TURBO_CACHE_DIR: .turbo/ios
     steps:
       - name: Checkout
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -154,6 +164,7 @@ jobs:
       - name: Install cocoapods
         if: env.turbo_cache_hit != 1 && steps.cocoapods-cache.outputs.cache-hit != 'true'
         run: |
+          yarn nitrogen
```
HayesGordon, line 167: Can decide this later, but might prefer to commit the nitrogen generated code

mfazekas: 👍  Yes let's do that, otherwise it's a pain with for example builder bob. Ideally we'd have a gh action that checks if the generated files are uptodate.

PR #2, opened 2025-10-28
chore: up nitro, up iOS (6.12.0), android(10.5.0), keep nitrogen generated in git

mfazekas:
- up nitro(0.31.3)
- up iOS (6.12.0)
- android(10.5.0) - required override in `androidx.core:core`, 
- keep nitrogen/generated in git

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "lint": "eslint \"**/*.{js,ts,tsx}\"",
     "clean": "del-cli android/build example/android/build example/android/app/build example/ios/build lib",
     "prepare": "yarn nitrogen && bob build",
-    "nitrogen": "nitro-codegen",
+    "nitrogen": "nitrogen",
     "release": "release-it",
     "dev:ios": "cd example && xed ios",
     "dev:android": "cd example && open -a \"/Applications/Android Studio.app\" ./android"
@@ -79,12 +79,12 @@
     "eslint-config-prettier": "^10.1.1",
     "eslint-plugin-prettier": "^5.2.3",
     "jest": "^29.7.0",
-    "nitro-codegen": "^0.25.2",
+    "nitrogen": "^0.31.3",
     "prettier": "^3.0.3",
     "react": "19.0.0",
     "react-native": "0.79.2",
     "react-native-builder-bob": "^0.40.10",
-    "react-native-nitro-modules": "^0.25.2",
+    "react-native-nitro-modules": "^0.31.3",
     "release-it": "^17.10.0",
     "turbo": "^1.10.7",
     "typescript": "^5.2.2"
--- a/src/specs/ViewModel.nitro.ts
+++ b/src/specs/ViewModel.nitro.ts
@@ -11,7 +11,7 @@ export interface ViewModel
   /** The number of view model instances in the view model */
   readonly instanceCount: number;
   /** The name of the view model */
-  readonly name: string;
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -79,12 +79,12 @@
     "eslint-config-prettier": "^10.1.1",
     "eslint-plugin-prettier": "^5.2.3",
     "jest": "^29.7.0",
-    "nitro-codegen": "^0.25.2",
+    "nitrogen": "^0.31.3",
```
HayesGordon, line 82: nice, i remember seeing somewhere that this name might change

src/specs/ViewModel.nitro.ts
```diff
@@ -11,7 +11,7 @@ export interface ViewModel
   /** The number of view model instances in the view model */
   readonly instanceCount: number;
   /** The name of the view model */
-  readonly name: string;
+  readonly modelName: string;
```
HayesGordon, line 14: Is `name` clashing with some other nitro generation?

src/specs/ViewModel.nitro.ts
```diff
@@ -52,9 +51,7 @@ export interface ViewModelInstance
 }
 
 export interface ViewModelProperty
-  extends HybridObject<{ ios: 'swift'; android: 'kotlin' }> {
-  dispose(): void;
```
HayesGordon, line 56: I believe some of this was me experimenting.

We will however need a way to clean up the native resources. I filed this issue on Nitro and looks like the request has already been implemented: https://github.com/mrousavy/nitro/issues/668

Something that we can investigate? It would be nice to clean up everything at once.

.github/workflows/ci.yml
```diff
@@ -23,8 +23,13 @@ jobs:
       - name: Lint files
         run: yarn lint
 
-      - name: Codegen
-        run: yarn nitrogen
+      - name: Validate Nitrogen Generated Code
+        run: |
+          yarn nitrogen
+          if ! git diff --exit-code nitrogen/generated/; then
+            echo "Error: nitrogen/generated/ is out of date. Please run 'yarn nitrogen' and commit the changes."
```
HayesGordon, line 30: Nice

src/specs/ViewModel.nitro.ts
mfazekas: Yes name is on HybridObject now. 

https://github.com/mrousavy/nitro/blob/67253832d1597796e3c32afa302e669fd54acb17/packages/react-native-nitro-modules/src/HybridObject.ts#L50-L53

And it's generated by nitrogen from the classname, so will be "ViewModel" for all ViewModels.
https://github.com/rive-app/rive-nitro-react-native/blob/ac840657301e80f560f3c68fb90823de404f3cdb/nitrogen/generated/ios/c%2B%2B/HybridViewModelSpecSwift.hpp#L40C1-L42C32

https://github.com/rive-app/rive-nitro-react-native/blob/ac840657301e80f560f3c68fb90823de404f3cdb/nitrogen/generated/shared/c%2B%2B/HybridViewModelSpec.hpp#L68

So I think it's best to use our own name methods for now.

src/specs/ViewModel.nitro.ts
mfazekas: Sure I'm looking into cleanup. The HOT reload issue is also related to cleanup, so I'll add to my todo list.

PR #3, opened 2025-10-29
refactor: use enums for aligment use union for properties

mfazekas:
- Use string based enums for alignment
- Use correct `properties?: Record<string, EventPropertiesOutput>;` instead of AnyMap
- Declare RiveEvent on TS side so that only type = RiveEventType.OpenUrl has url and target fields.

Marc  suggested to use different listeners for each type. So there would be an `onURLEvent` on `onGeneralEvent`, but it's a bit of a pain as we need to maintain listeners for that and remove functions, etc for every event type.  But we could convert back to a single function on TS side. We can covert to that solution later.


src/core/Events.ts
```diff
@@ -1,42 +1,31 @@
...
   type: RiveEventType;
   delay?: number;
-  // properties?: Record<string, EventPropertiesOutput>; // TODO: This is failing on Nitro android
-  properties?: AnyMap;
+  properties?: Record<string, EventPropertiesOutput>;
   url?: string;
   target?: string;
 }
+
+export type RiveEvent = {
```
HayesGordon, line 17: This is really nice. I always forget TypeScript can do this.

ios/EventPropertiesOutputExtensions.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+extension EventPropertiesOutput {
```
HayesGordon, line 3: Nice

PR #4, opened 2025-10-29
chore: add support to customize ios and android sdk versions

mfazekas:
Adds support for customizing Rive native SDK versions on  iOS and Android.

- default native sdk versions moved to `package.json` (`runtimeVersions.ios` and `runtimeVersions.android`)
- allow users to customize the versions. In `ios/Podfile.properties.json`set `"RiveRuntimeIOSVersion":"6.13.0"` key. In `android/gradle.properties` `Rive_RiveRuntimeAndroidVersion=10.6.0`

### DIFF
--- a/Rive.podspec
+++ b/Rive.podspec
@@ -2,6 +2,34 @@ require "json"
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
 
+rive_ios_version = nil
+
+if ENV['RIVE_RUNTIME_IOS_VERSION']
+  rive_ios_version = ENV['RIVE_RUNTIME_IOS_VERSION']
+end
+
+if !rive_ios_version && defined?($RiveRuntimeIOSVersion)
+  rive_ios_version = $RiveRuntimeIOSVersion
+end
+
+if !rive_ios_version && defined?(Pod::Config) && Pod::Config.respond_to?(:instance)
+  podfile_properties_path = File.join(Pod::Config.instance.installation_root, 'Podfile.properties.json')
+  if File.exist?(podfile_properties_path)
+    podfile_properties = JSON.parse(File.read(podfile_properties_path)) rescue {}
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -59,6 +59,10 @@
     "url": "https://github.com/rive-app/rive-nitro-react-native/issues"
   },
   "homepage": "https://github.com/rive-app/rive-nitro-react-native#readme",
+  "runtimeVersions": {
```
HayesGordon, line 62: So nice

android/build.gradle
```diff
@@ -122,10 +122,33 @@ repositories {
 
 def kotlin_version = getExtOrDefault("kotlinVersion")
 
+def getRiveAndroidVersion() {
+  if (rootProject.ext.has('RiveRuntimeAndroidVersion')) {
```
HayesGordon, line 126: When/how will this be set? Should we document this option as well?

I see on iOS we also have an environment option. Does Android need the same?

Rive.podspec
```diff
@@ -2,6 +2,34 @@ require "json"
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
 
+rive_ios_version = nil
+
+if ENV['RIVE_RUNTIME_IOS_VERSION']
```
HayesGordon, line 7: Should we document the environment option as well?

HayesGordon: Or alternatively simplify and only support one option (for now)

android/build.gradle
mfazekas: I've looked into various libraries, some use this ext method some env variables. `iOS/Podfile.properties.json`, `android/gradle.properties` sounds the most straightforward to me considering expo, I've documented only that.

Not sure we need all 3. What do you think?

HayesGordon: No, I also don't think we need all the options. But is there then an argument to remove the option completely. Instead of just not documenting?

PR #12, opened 2025-10-31
fix: iOS production builds fail to load bundled .riv assets

mfazekas:
## Problem

iOS production builds fail to load bundled `.riv` assets with error:
```
RiveFileFactory.fromResource(...): Could not find Rive file: rewards_source.riv
```

image

### Root Cause

On iOS, `resolveAssetSource()` returns file:// URLs for bundled assets:
```
file:///Users/.../RiveExample.app/assets/assets/rive/rewards_source.riv
```

**The previous implementation:**
- Extracted the filename using regex: `rewards_source`
- Tried to load via `fromResource("rewards_source")`
- This worked in **debug builds** where assets url-s to the bundler
- **Failed in production builds** where assets are packaged as files

**Why it failed in production:**
In production, assets exist as physical files in the app bundle deeper, not as named resources accessible via `Bundle.main.path(forResource:)`.

## Solution

Load bundled assets directly from their file:// URLs instead of trying to extract and load as bundle resources. To support this fix, added `fromFileURL()` method to `RiveFileFactory`:

### DIFF
--- a/src/core/RiveFile.ts
+++ b/src/core/RiveFile.ts
@@ -31,7 +31,7 @@ export namespace RiveFileFactory {
 
   /**
    * Creates a RiveFile instance from a local file path URL.
-   * @param pathURL - The local file path of the Rive animation file
+   * @param pathURL - The local file path of the Rive graphic file
    * @param loadCdn - Whether to load from CDN (default: true)
    * @returns Promise that resolves to a RiveFile instance
    */


src/core/RiveFile.ts
```diff
@@ -29,6 +29,19 @@ export namespace RiveFileFactory {
     return RiveFileInternal.fromURL(url, loadCdn);
   }
 
+  /**
+   * Creates a RiveFile instance from a local file path URL.
+   * @param pathURL - The local file path of the Rive animation file
+   * @param loadCdn - Whether to load from CDN (default: true)
```
HayesGordon, line 35: The option for `loadCdn` reminds me that we should revisit this. I believe Android doesn't provide the option. For now we can opt to remove this and stick to the defaults. I'll add a ticket to revisit

src/core/RiveFile.ts
```diff
@@ -29,6 +29,19 @@ export namespace RiveFileFactory {
     return RiveFileInternal.fromURL(url, loadCdn);
   }
 
+  /**
+   * Creates a RiveFile instance from a local file path URL.
+   * @param pathURL - The local file path of the Rive animation file
```
HayesGordon:
```suggestion
   * @param pathURL - The local file path of the Rive graphic file
```

src/specs/RiveFile.nitro.ts
```diff
@@ -22,6 +22,7 @@ export interface RiveFile
 export interface RiveFileFactory
   extends HybridObject<{ ios: 'swift'; android: 'kotlin' }> {
   fromURL(url: string, loadCdn: boolean): Promise<RiveFile>;
+  fromFileURL(fileURL: string, loadCdn: boolean): Promise<RiveFile>;
```
HayesGordon, line 25: This is great!

src/core/RiveFile.ts
HayesGordon: I know the other areas refer to it as "animation" - but "animation" no longer encapsulates everything Rive can do. I'll create an issue for this too, so that we refer to Rive as a graphics format, instead of just animations in the docs/API

Issue #7, opened 2025-10-31
Add referenced (out-of-band) asset support


HayesGordon:
See [Loading Assets](https://rive.app/docs/runtimes/loading-assets).

This package needs similar support to load in (and update) referenced assets.

PR #16, opened 2025-11-06
feat: referenced assets

mfazekas:
Implements referenced assets for RiveFile on iOS/Android. Create and update.

Since it's on file level to refresh views we need week links to views using that file. Also there is synchronisation so we only refresh after asset download and update has finished. Otherwise it's the same as the non nitro version.

Fixes: #7

### DIFF
--- a/ios/HybridRiveFileFactory.swift
+++ b/ios/HybridRiveFileFactory.swift
@@ -4,46 +4,6 @@ import RiveRuntime
 final class HybridRiveFileFactory: HybridRiveFileFactorySpec, @unchecked Sendable {
   let assetLoader = ReferencedAssetLoader()
 
-  private final func buildRiveFile(
-    fileWithCustomAssetLoader: (@escaping LoadAsset) throws -> RiveFile,
-    file: () throws -> RiveFile,
-    referencedAssets: ReferencedAssetsType?
-  ) throws -> (
-    file: RiveFile, cache: ReferencedAssetCache, factory: RiveFactory?,
-    loader: ReferencedAssetLoader?
-  ) {
-    let referencedAssetCache = SendableRef(ReferencedAssetCache())
-    let factoryCache: SendableRef<RiveFactory?> = .init(nil)
-    let customLoader = assetLoader.createCustomLoader(
-      referencedAssets: referencedAssets, cache: referencedAssetCache, factory: factoryCache)
-
-    let riveFile =
-      if let customLoader = customLoader {
-        try fileWithCustomAssetLoader(customLoader)
-      } else {
-        try file()
-      }
-
-    return (
-      file: riveFile, cache: referencedAssetCache.value, factory: factoryCache.value,
-      loader: customLoader != nil ? assetLoader : nil
-    )
-  }
-
-  private func buildRiveFile(data: Data, loadCdn: Bool, referencedAssets: ReferencedAssetsType?)
-    throws -> (
-      file: RiveFile, cache: ReferencedAssetCache, factory: RiveFactory?,
-      loader: ReferencedAssetLoader?
-    )
-  {
-    return try buildRiveFile(
-      fileWithCustomAssetLoader: { (loader) in
-        try RiveFile(data: data, loadCdn: loadCdn, customAssetLoader: loader)
-      },
-      file: { try RiveFile(data: data, loadCdn: loadCdn) },
-      referencedAssets: referencedAssets)
-  }
-
   /// Asynchronously creates a `HybridRiveFileSpec` by performing the following steps:
   /// 1. Executes `check()` to validate or fetch initial data.
   /// 2. Processes the result with `prepare()`.


src/hooks/useRiveFile.ts
```diff
@@ -1,50 +1,146 @@
...
+  }
+
+  Object.entries(mapping).forEach(([key, option]) => {
+    transformedMapping[key] = parsePossibleSources(option.source);
+  });
+
+  return transformedMapping;
+}
+
+type RiveFileHookResult =
```
HayesGordon, line 70: nice

src/hooks/useRiveFile.ts
```diff
@@ -1,50 +1,146 @@
...
+  | { riveFile: RiveFile; isLoading: false; error: null }
+  | { riveFile: null; isLoading: true; error: null }
+  | { riveFile: null; isLoading: false; error: string };
+
+export function useRiveFile(
+  input: RiveFileInput | undefined,
```
HayesGordon, line 76: What is the benefit of allowing undefined?

mfazekas: I think RiveFileExample used that for example. Basically hook composability, is much easier if we allow undefined.

```tsx
const [input, setInput] = useState<RiveFileInput | undefined>(undefined);
```

https://github.com/rive-app/rive-nitro-react-native/blob/09e0faee9359a297207747962d58e901f272fd67/example/src/pages/RiveFileLoadingExample.tsx#L54-L71

src/hooks/__tests__/useRiveFile.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    (global as any).mockRiveFileFactory.fromURL.mockResolvedValue(mockRiveFile);
+    (global as any).mockRiveFileFactory.fromResource.mockResolvedValue(
+      mockRiveFile
+    );
+  });
+
+  it('should call updateReferencedAssets when referencedAssets change', async () => {
```
HayesGordon, line 29: Whoop! Tests!

src/index.tsx
```diff
@@ -50,6 +50,7 @@ export const RiveView = getHostComponent<RiveViewProps, RiveViewMethods>(
   'RiveView',
   () => RiveViewConfig
 ) as ReactNativeView<RiveViewProps, RiveViewTSMethods>;
+export { RiveView as NitroRiveView };
```
HayesGordon: Is this needed for this PR, or for that other add on PR you made?

ios/HybridRiveView.swift
```diff
@@ -83,19 +83,21 @@ class HybridRiveView : HybridRiveViewSpec {
...
+    guard let hybridFile = file as? HybridRiveFile,
+          let file = hybridFile.riveFile else { return }
+
     let config = ViewConfiguration(
       artboardName: artboardName,
       stateMachineName: stateMachineName,
       autoBind: autoBind ?? DefaultConfiguration.autoBind,
       autoPlay: autoPlay ?? DefaultConfiguration.autoPlay,
       riveFile: file,
+      hybridRiveFile: hybridFile,
```
HayesGordon: I initially wanted to decouple nitro/hybrid logic from the Rive view - or to at least keep that boundary as closed down as possible.

I wonder if we could register the weak view here instead?

Alternatively, if that can't be avoided, then it seems like we can instead just directly pass `hybridFile` and extract `riveFile` in the view?

ios/HybridRiveFileFactory.swift
```diff
@@ -26,35 +41,84 @@ class HybridRiveFileFactory: HybridRiveFileFactorySpec {
...
+        let factoryCache = SendableRef<RiveFactory?>(nil)
+        let riveFile = try await withCheckedThrowingContinuation { continuation in
+          DispatchQueue.global(qos: .userInitiated).async {
+            do {
+              let riveFile =
+              if let customLoader = assetLoader.createCustomLoader(referencedAssets: referencedAssets, cache: referencedAssetCache, factory: factoryCache) {
+                  try RiveFile(resource: resource, loadCdn: loadCdn, customAssetLoader: customLoader)
+                } else {
+                  try RiveFile(resource: resource, loadCdn: loadCdn)
+                }
```
HayesGordon: Can this also use `buildRiveFile` like the other methods?

ios/HybridRiveFileFactory.swift
```diff
@@ -128,14 +200,18 @@ class HybridRiveFileFactory: HybridRiveFileFactorySpec {
             }
           }
         }
-        
+
         let hybridRiveFile = HybridRiveFile()
-        hybridRiveFile.riveFile = riveFile
+        hybridRiveFile.riveFile = result.file
+        hybridRiveFile.referencedAssetCache = result.cache
+        hybridRiveFile.assetLoader = result.loader
```
HayesGordon: We're repeating this logic in a few places. Something that happened with the old runtime is that something would work through method x, but not method y.

Could we modify `buildRiveFile` to take in the `hybridRiveFile` and do all of these assignments there?

src/index.tsx
mfazekas: Good catch I'm removing it.

ios/RiveReactNativeView.swift
```diff
@@ -67,7 +74,13 @@ class RiveReactNativeView: UIView, RiveStateMachineDelegate {
   func pause() {
     baseViewModel?.pause()
   }
-  
+
+  func refreshAfterAssetChange() {
```
HayesGordon, line 83: We could make something a bit more generic. `refresh` or `restart`.

The reason is because atm Android and iOS also does not restart the underlying graphic if it's in a settled state when updating data binding.

Something we should chat about though as part of a separate PR. Additionally, we should introduce the concept of `paused/stopped/active` in ReactNative that works independently from calling `play`/`pause`. But this may need to wait for the iOS and Android rewrites.

ios/ReferencedAssetLoader.swift
```diff
@@ -0,0 +1,235 @@
...
+    guard let referencedAssets = referencedAssets, let referencedAssets = referencedAssets.data
+    else {
+      return nil
+    }
+    return { (asset: RiveFileAsset, data: Data, factory: RiveFactory) -> Bool in
+      let assetByUniqueName = referencedAssets[asset.uniqueName()]
+      guard let assetData = assetByUniqueName ?? referencedAssets[asset.name()] else {
+        return false
+      }
+      let usedKey = assetByUniqueName != nil ? asset.uniqueName() : asset.name()
```
HayesGordon: This is unused. But does make me question whether we should always be using `uniqueName` here instead of this key

ios/ReferencedAssetLoader.swift
```diff
@@ -0,0 +1,235 @@
...
+
+      cache.value[asset.uniqueName()] = asset
+      factoryOut.value = factory
+
+      self.loadAssetInternal(source: assetData, asset: asset, factory: factory, completion: {})
+
+      return false
```
HayesGordon: This should be `true`. Indicating that Rive should not try to load in the asset if it is present - typically it won't be to benefit from refernced assests.

ios/HybridRiveView.swift
mfazekas: We can add a protocol there to decouple

example/src/pages/OutOfBandAssets.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const { riveFile, isLoading, error } = useRiveFile(
+    require('../../assets/rive/out_of_band.riv'),
+    {
+      referencedAssets: {
+        'Inter-594377': {
+          source: require('../../assets/fonts/Inter-594377.ttf'),
+          // source: {
+          //   fileName: 'Inter-594377.ttf',
+          //   path: 'fonts', // only needed for Android assets
+          // },
```
HayesGordon, line 28: We should add this (and the other two fileName + path options to the Android and iOS resources) if we want to keep these as examples

src/hooks/useRiveFile.ts
HayesGordon: Makes sense!

ios/HybridRiveFileFactory.swift
HayesGordon: Or instead of taking it in (after looking at Android) it can return a `HybridRiveFile`

ios/HybridRiveView.swift
mfazekas: Added protocol called `RiveViewSource`, a naming could be probably better. 

```
protocol RiveViewSource: AnyObject {
  func registerView(_ view: RiveReactNativeView)
  func unregisterView(_ view: RiveReactNativeView)
}
```

ios/HybridRiveFileFactory.swift
mfazekas: refactored to use a single function on all

ios/HybridRiveFileFactory.swift
```diff
@@ -1,23 +1,98 @@
...
 import RiveRuntime
 
-class HybridRiveFileFactory: HybridRiveFileFactorySpec {
-  // MARK: Public Methods
-  func fromURL(url: String, loadCdn: Bool) throws -> Promise<(any HybridRiveFileSpec)> {
-    // TODO: should we make use of the underlying Rive iOS URL asset loading instead
+final class HybridRiveFileFactory: HybridRiveFileFactorySpec, @unchecked Sendable {
+  let assetLoader = ReferencedAssetLoader()
+
+  private final func buildRiveFile(
```
HayesGordon: I'm missing where this is used now?

mfazekas: my bad, was removed now

PR #20, opened 2025-11-13
chore: add an expo based example as well

mfazekas:
- added an expo based example app  (uses example from example/src/pages/OutOfBandAssets.tsx)
- in the expo app iOS and android is ignored, so it's easy to upgrade

### DIFF
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -1,4 +1,4 @@
 {
   "extends": "./tsconfig",
-  "exclude": ["example", "lib"]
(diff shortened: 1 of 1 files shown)


expo-example/README.md
```diff
@@ -0,0 +1,50 @@
+# Welcome to your Expo app 👋
```
HayesGordon, line 1: Any specific Rive instructions to add in this example README?

mfazekas: I'll revise readme in another pr

PR #23, opened 2025-11-13
feat: add dataBind prop to RiveView

mfazekas:
Add a dataBind property to allow binding viewModelInstance in a declarative way.

Usages:
1.) Set and instance => no ref needed
   
   ```tsx
   const { riveViewRef, setHybridRef } = useRive();
   
    useEffect(() => {
      if (riveViewRef) {
        riveViewRef.bindViewModelInstance(instance);
      }
    }, [riveViewRef, instance]);
    
    
    return (
      <RiveView
         hybridRef={setHybridRef}
       />
    )
   ```
   
    to just:
    
   ```tsx
    return (
      <RiveView
         bind={instance}
       />
    )
   ```
   
2.) Also accepts by name:
  
  ```tsx
  <RiveView dataBind={{byName: 'green'}} />
  ``` 

3.) None or Auto for default
  ```tsx
  <RiveView dataBind={DataBindMode.AUTO} />
  ```

### DIFF
--- a/android/src/main/java/com/margelo/nitro/rive/HybridRiveView.kt
+++ b/android/src/main/java/com/margelo/nitro/rive/HybridRiveView.kt
@@ -50,6 +50,7 @@ class HybridRiveView(val context: ThemedReactContext) : HybridRiveViewSpec() {
   override val view: RiveReactNativeView = RiveReactNativeView(context)
   private var needsReload = false
   private var dataBindingChanged = false
+  private var initialUpdate = true
   private var registeredFile: HybridRiveFile? = null
   //endregion
 
@@ -156,7 +157,7 @@ class HybridRiveView(val context: ThemedReactContext) : HybridRiveViewSpec() {
         layoutScaleFactor = layoutScaleFactor?.toFloat() ?: DefaultConfiguration.LAYOUTSCALEFACTOR,
         bindData = dataBind.toBindData()
       )
-      view.configure(config, dataBindingChanged=dataBindingChanged, needsReload)
+      view.configure(config, dataBindingChanged=dataBindingChanged, needsReload, initialUpdate= initialUpdate)
 
       if (needsReload && hybridFile != null) {
         hybridFile.registerView(this)
@@ -165,6 +166,7 @@ class HybridRiveView(val context: ThemedReactContext) : HybridRiveViewSpec() {
 
       needsReload = false
       dataBindingChanged = false
+      initialUpdate = false
       super.afterUpdate()
     }
   }
--- a/android/src/main/java/com/rive/RiveReactNativeView.kt
+++ b/android/src/main/java/com/rive/RiveReactNativeView.kt
@@ -57,7 +57,7 @@ class RiveReactNativeView(context: ThemedReactContext) : FrameLayout(context) {
     return viewReadyDeferred.await()
   }
 
-  fun configure(config: ViewConfiguration, dataBindingChanged: Boolean, reload: Boolean = false) {
+  fun configure(config: ViewConfiguration, dataBindingChanged: Boolean, reload: Boolean = false, initialUpdate: Boolean = false) {
     if (reload) {
       riveAnimationView?.setRiveFile(
         config.riveFile,
@@ -76,7 +76,7 @@ class RiveReactNativeView(context: ThemedReactContext) : FrameLayout(context) {
       riveAnimationView?.layoutScaleFactor = config.layoutScaleFactor
     }
 
-    if (dataBindingChanged) {
+    if (dataBindingChanged || initialUpdate) {
       applyDataBinding(config.bindData)
     }
 


expo-example/app.json
```diff
@@ -20,7 +20,8 @@
         "monochromeImage": "./assets/images/android-icon-monochrome.png"
       },
       "edgeToEdgeEnabled": true,
-      "predictiveBackGestureEnabled": false
+      "predictiveBackGestureEnabled": false,
+      "package": "com.mfazekas.expoexample"
```
HayesGordon: Doesn't need to be this PR, but can we identify this package to have the rive domain

HayesGordon: Same for other areas

example/src/pages/RiveTextRunExample.tsx
```diff
@@ -31,7 +37,7 @@ export default function TextRunExample() {
         ) : riveFile ? (
           <RiveView
             style={styles.rive}
-            autoBind={false}
+            dataBind={DataBindMode.None}
```
HayesGordon: Could exclude from showing this? Now that Auto will handle none if the file doesn't have DataBinding. The two features will work together, but data binding is just better than setting text this way.

example/src/pages/RiveStateMachineInputsExample.tsx
```diff
@@ -47,7 +53,7 @@ export default function StateMachineInputsExample() {
         ) : riveFile ? (
           <RiveView
             style={styles.rive}
-            autoBind={false}
+            dataBind={DataBindMode.None}
```
HayesGordon: could exclude?

example/src/pages/RiveEventsExample.tsx
```diff
@@ -46,7 +47,7 @@ export default function EventsExample() {
         ) : riveFile ? (
           <RiveView
             style={styles.rive}
-            autoBind={false}
+            dataBind={DataBindMode.None}
```
HayesGordon: could exclude?

example/src/pages/RiveDataBindingExample.tsx
```diff
@@ -104,12 +95,11 @@ function DataBindingExample({
   return (
     <RiveView
       style={styles.rive}
-      autoBind={false}
       autoPlay={true}
+      dataBind={instance}
```
HayesGordon, line 99: 🥳

example/src/pages/OutOfBandAssets.tsx
```diff
@@ -65,6 +66,7 @@ export default function StateMachine() {
         fit={Fit.Contain}
         style={styles.animation}
         stateMachineName="State Machine 1"
+        dataBind={DataBindMode.None}
```
HayesGordon: exclude?

example/src/pages/RiveFileLoadingExample.tsx
```diff
@@ -81,7 +82,7 @@ const CustomRiveView = ({ loadingMethod, title }: CustomRiveViewProps) => {
       ) : riveFile ? (
         <RiveView
           style={styles.rive}
-          autoBind={false}
+          dataBind={DataBindMode.None}
```
HayesGordon: exclude?

ios/RiveReactNativeView.swift
```diff
@@ -66,12 +74,51 @@ class RiveReactNativeView: UIView, RiveStateMachineDelegate {
...
+    let stateMachine = baseViewModel?.riveModel?.stateMachine
+    let artboard = baseViewModel?.riveModel?.artboard
+
+    switch bindData {
+    case .none:
+      baseViewModel?.riveModel?.disableAutoBind()
+      
+
+    case .auto:
+      baseViewModel?.riveModel?.enableAutoBind { [weak self] instance in
```
HayesGordon:
```suggestion
      baseViewModel?.riveModel?.enableAutoBind { instance in
```

ios/RiveReactNativeView.swift
```diff
@@ -66,12 +74,51 @@ class RiveReactNativeView: UIView, RiveStateMachineDelegate {
...
+
+    case .byName(let name):
+      guard let artboard = artboard,
+            let riveFile = baseViewModel?.riveModel?.riveFile,
+            let viewModel = riveFile.defaultViewModel(for: artboard),
+            let instance = viewModel.createInstance(fromName: name) else {
+        return
+      }
+      stateMachine?.bind(viewModelInstance: instance)
+      artboard.bind(viewModelInstance: instance)
```
HayesGordon: Might need to keep an eye on `artboard.bind`. We originally needed it, but now internally if we call `stateMachine.bind` we also bind the artboard.

Technically you can just play an artboard and bind data. But not sure if iOS (or Android) allows just playing an artboard.

I'd say let's comment it out for now with a note saying that this should be added if we support only playing artboards on their own in the future.

ios/HybridRiveView.swift
```diff
@@ -1,77 +1,116 @@
...
+      }
+    case .third(let dataBindByName):
+      return .byName(dataBindByName.byName)
+    }
+  }
+}
+
+class HybridRiveView: HybridRiveViewSpec {
+  var firstUpdate = true
+  var dataBind: Variant__any_HybridViewModelInstanceSpec__DataBindMode_DataBindByName? = nil {
```
HayesGordon: ouch, long name

ios/Utils.swift
```diff
@@ -13,3 +13,12 @@ final class SendableRef<T: Sendable>: @unchecked Sendable {
     self.value = value
   }
 }
+
+/// Executes a closure, logging any thrown error with an optional note and tag using RCTLogError.
+func logged(tag: String, note: String? = nil, _ fn: () throws -> Void) {
```
HayesGordon, line 18: Nice, one thing to consider is how users will handle these view errors in production. If we wanted to give a way to listen to errors that are produced? Not sure what the React Native way is for this.

ios/HybridRiveView.swift
```diff
@@ -1,77 +1,116 @@
...
+        return .none
+      }
+    case .third(let dataBindByName):
+      return .byName(dataBindByName.byName)
+    }
+  }
+}
+
+class HybridRiveView: HybridRiveViewSpec {
+  var firstUpdate = true
```
HayesGordon: Can we move these two to MARK: Internal state down below.

ios/RiveReactNativeView.swift
```diff
@@ -66,12 +74,51 @@ class RiveReactNativeView: UIView, RiveStateMachineDelegate {
...
+
+    case .instance(let instance):
+      stateMachine?.bind(viewModelInstance: instance)
+      artboard?.bind(viewModelInstance: instance)
+    }
+    if (refresh) {
+      baseViewModel?.play()
```
HayesGordon: Did you see issues when calling `play` when it's already playing? If yes, then it's something we should resolve at the native iOS side @dskuza.

Ideally we can simplify the logic and just call play without needing to track this conditional

android/src/main/java/com/rive/RiveReactNativeView.kt
```diff
@@ -77,6 +87,55 @@ class RiveReactNativeView(context: ThemedReactContext) : FrameLayout(context) {
...
+
+    when (bindData) {
+      is BindData.None -> {
+        stateMachine.viewModelInstance = null
+      }
+      is BindData.Auto -> {
+        val artboard = riveAnimationView?.controller?.activeArtboard
+        val file = riveAnimationView?.controller?.file
+        if (artboard != null && file != null) {
+          val viewModel = file.defaultViewModelForArtboard(artboard)
```
HayesGordon: This is a throwable function. We need to try catch this one. If it throws then there is no databinding, and auto should function as None in that scenario (to cater for old files that does not have data binding)

HayesGordon: same for below. In general android can throw for many/all of these data binding operations.

android/src/main/java/com/margelo/nitro/rive/HybridRiveView.kt
```diff
@@ -103,24 +135,31 @@ class HybridRiveView(val context: ThemedReactContext) : HybridRiveViewSpec() {
...
       stateMachineName = stateMachineName,
       autoPlay = autoPlay ?: DefaultConfiguration.AUTOPLAY,
-      autoBind = autoBind ?: DefaultConfiguration.AUTOBIND,
       riveFile = riveFile,
       alignment = convertAlignment(alignment) ?: DefaultConfiguration.ALIGNMENT,
       fit = convertFit(fit) ?: DefaultConfiguration.FIT,
       layoutScaleFactor = layoutScaleFactor?.toFloat() ?: DefaultConfiguration.LAYOUTSCALEFACTOR,
+      bindData = dataBind.toBindData()
     )
     view.configure(config, needsReload)
```
HayesGordon: Should we have similar logs to what you implemented on iOS?

android/src/main/java/com/rive/RiveReactNativeView.kt
mfazekas: In the AUTO case we should silently ignore the error, in the ByName case we should treat it as an error, and log it and call onError if there is one.

ios/Utils.swift
mfazekas: In @rnmapbox/maps for example I have a singleton logger module, where user can add callback on the js side and logs are either suppressed, or process logs in production.

But it think this should be going to the onError callback of the RiveView, which is to be implemented

expo-example/app.json
```diff
@@ -20,7 +20,8 @@
         "monochromeImage": "./assets/images/android-icon-monochrome.png"
       },
       "edgeToEdgeEnabled": true,
-      "predictiveBackGestureEnabled": false
+      "predictiveBackGestureEnabled": false,
+      "package": "com.rive.expoexample"
```
HayesGordon, line 24: Note the iOS buildIdentifier above is still `com.mfazekas.`

expo-example/app.json
mfazekas:
```suggestion
      "bundleIdentifier": "com.rive.expo-example"
```

PR #25, opened 2025-11-19
feat: onError property for RiveView

mfazekas:
Introduce `onError()` callback on `RiveView`. This is a required fn and defaults to `console.error` 

```tsx
onError(error: {message: string, type: RiveErrorType})


```

image


src/core/Errors.ts
```diff
@@ -0,0 +1,15 @@
+export enum RiveErrorType {
+  Unknown = 0,
+  FileNotFound = 1,
+  MalformedFile = 2,
+  IncorrectArtboardName = 3,
+  IncorrectStateMachineName = 4,
+  IncorrectAnimationName = 5,
```
HayesGordon: This one can be removed. This version only supports playing state machines

ios/HybridRiveView.swift
```diff
@@ -187,3 +188,52 @@ class HybridRiveView: HybridRiveViewSpec {
...
+      return .incorrectstatemachinename
+    case "NoAnimationFound":
+      return .incorrectanimationname
+    case "Malformed":
+      return .malformedfile
+    case "FileNotFound":
+      return .filenotfound
+    case "NoStateMachineInputFound":
+      return .incorrectanimationname
+    case "TextRunNotFoundError":
```
HayesGordon: Looks like this is not in the iOS runtime anymore.. or maybe was something special added to the old RN runtime

ios/HybridRiveView.swift
```diff
@@ -187,3 +188,52 @@ class HybridRiveView: HybridRiveViewSpec {
...
+      return .textrunnotfounderror
+    case "DataBindingError":
```
HayesGordon: Same for this. Not at the iOS level, so maybe something unique added to the old RN runtime

ios/HybridRiveView.swift
```diff
@@ -187,3 +188,52 @@ class HybridRiveView: HybridRiveViewSpec {
...
+    switch errorName {
+    case "NoArtboardFound":
+      return .incorrectartboardname
+    case "NoStateMachineFound":
+      return .incorrectstatemachinename
+    case "NoAnimationFound":
+      return .incorrectanimationname
+    case "Malformed":
+      return .malformedfile
+    case "FileNotFound":
```
HayesGordon: Also not in the iOS runtime

PR #36, opened 2025-11-21
chore: add example view problem

HayesGordon:
Looks like we have a view cleanup (rendering) bug on Android when transitioning between pages. It doesn't happen every time, though. See the video where it only happened after trying a couple of times.

At the current time, this PR is just extending one of the examples to show how to reproduce the issue, not resolving anything.

https://github.com/user-attachments/assets/9392e1ee-6b0e-46ca-a53f-4a528b6d9cfe


mfazekas:
This was fixed

Issue #9, opened 2025-10-31
Add data binding images support


HayesGordon:
See [Data Binding images](https://rive.app/docs/runtimes/data-binding#images)

PR #40, opened 2025-11-25
feat: ViewModelImageProperty

mfazekas:
Add Image type data binding - set only

Fixes: #9

### DIFF
--- a/android/src/main/java/com/margelo/nitro/rive/HybridViewModelImageProperty.kt
+++ b/android/src/main/java/com/margelo/nitro/rive/HybridViewModelImageProperty.kt
@@ -15,7 +15,7 @@ class HybridViewModelImageProperty(private val viewModelImage: ViewModelImagePro
   }
 
   override fun addListener(onChanged: () -> Unit) {
-    listeners.add { onChanged () }
+    listeners.add(onChanged)
     ensureValueListenerJob(viewModelImage.valueFlow.map { })
   }
 }


example/src/pages/ManyViewModels.tsx
```diff
@@ -91,13 +97,60 @@ export default function ManyViewModels() {
...
+        return;
+      }
+
+      const riveImage = await RiveImages.loadFromURLAsync(
+        'https://picsum.photos/id/372/500/500'
+      );
+      imgProp.set(riveImage);
+      riveViewRef.current.play();
+
+      imgProp.addListener(() => {
```
HayesGordon: Might be bad practice showing adding a listener and not removing it. While playing around with the example I noticed that we'd end up getting multiple calls if you spam the button.

ios/HybridViewModelImageProperty.swift
```diff
@@ -0,0 +1,46 @@
+import RiveRuntime
+
+class HybridViewModelImageProperty: HybridViewModelImagePropertySpec {
```
HayesGordon: I think we can update this to use the newly added `ValuedPropertyProtocol`

ios/BaseHybridViewModelProperty.swift
```diff
@@ -4,8 +4,15 @@ import RiveRuntime
...
-  
-  func addListener(_ callback: @escaping (ListenerValueType) -> Void) -> UUID
+  typealias ListenerType = (ListenerValueType) -> Void
+
+  func addListener(_ callback: @escaping ListenerType) -> UUID
+  func removeListener(_ id: UUID)
+}
+
+/// Protocol for Rive property types with void listeners (Trigger, Image)
+protocol RivePropertyWithVoidListeners: AnyObject {
```
HayesGordon, line 14: nice

PR #43, opened 2025-11-25
fix(artboard): setting the artboard and statemachine

bassrock:
## Description

Ensuring the statemachine and artboard are passed in


mfazekas:
@bassrock thank much for the pr, looks good! 

Can you please describe in a sentence or two the issue you faced and the PR solves, thanks

bassrock:
@mfazekas We use a single Rive File for some work, in some cases we need to change which artboard is displayed for the user. It worked great on Android, but on iOS only the default artboard was ever showing. iOS had a bug where it was not passing in the configured artboard or statemachine.

mfazekas (approved):

HayesGordon:
Thanks for the contribution!

PR #55, opened 2025-11-28
refactor(ios): shared property listener implementation

HayesGordon:
This might be overkill. But I noticed in this [PR](https://github.com/rive-app/rive-nitro-react-native/pull/40), how much nicer the Kotlin code is.

I tried back in the day to abstract the Swift code but the properties don't inherit from the same base. This is another attempt by adding protocol extensions + composition with a helper class.

### DIFF
--- a/ios/BaseHybridViewModelProperty.swift
+++ b/ios/BaseHybridViewModelProperty.swift
@@ -66,18 +66,17 @@ class PropertyListenerHelper<PropertyType: RivePropertyWithListeners> {
 
 /// Protocol for properties that have typed values (Bool, String, Double, etc.)
 /// Provides a default addListener implementation
-protocol ValuedPropertyProtocol<ValueType>{
+protocol ValuedPropertyProtocol<ValueType> {
   associatedtype PropertyType: RivePropertyWithListeners
   associatedtype ValueType
   
   var property: PropertyType! { get }
   var helper: PropertyListenerHelper<PropertyType> { get }
-
+  
   func addListener(onChanged: @escaping (ValueType) -> Void) throws
   func removeListeners() throws
   func dispose() throws
 }
-
 
 /// Default implementations for lifecycle methods (always available)
 extension ValuedPropertyProtocol {
@@ -96,4 +95,3 @@ extension ValuedPropertyProtocol where PropertyType.ListenerValueType == ValueTy
     helper.addListener(onChanged)  // Types match, just forward directly!
   }
 }
-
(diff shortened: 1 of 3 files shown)


ios/HybridViewModelBooleanProperty.swift
```diff
@@ -28,20 +28,10 @@ class HybridViewModelBooleanProperty: HybridViewModelBooleanPropertySpec {
...
-    let id = property.addListener({ value in
-      onChanged(value)
-    })
-    listenerIds.append(id)
-  }
-  
-  func removeListeners() throws {
-    for id in listenerIds {
-      property.removeListener(id)
+    helper.trackListener { property in
```
mfazekas: Nice! 

Btw can't we make a helper.addListener?! that does call trackListener and property.addListener?!

mfazekas (approved):
💯

ios/HybridViewModelBooleanProperty.swift
HayesGordon: I'm cleaning this up!

ios/BaseHybridViewModelProperty.swift
```diff
@@ -0,0 +1,99 @@
...
+    try helper.removeListeners()
+  }
+  
+  func dispose() throws {
+    try helper.dispose()
+  }
+}
+
+/// Automatic addListener() ONLY when ListenerValueType == ValueType (no conversion needed)
+extension ValuedPropertyProtocol where PropertyType.ListenerValueType == ValueType {
```
HayesGordon, line 93: We now automatically also handle adding a listener if the native type and the converted type is the same. If not, you have to provide the `addListener` implementation.

ios/BaseHybridViewModelProperty.swift
```diff
@@ -0,0 +1,99 @@
...
+extension StringPropertyType: RivePropertyWithListeners {
+  typealias ListenerValueType = String  // Native: String → String (no conversion)
+}
+extension EnumPropertyType: RivePropertyWithListeners {
+  typealias ListenerValueType = String  // Native: String → String (no conversion)
+}
+extension ColorPropertyType: RivePropertyWithListeners {
+  typealias ListenerValueType = UIColor  // Native: UIColor → Double (needs conversion)
+}
+// Note: TriggerProperty doesn't fit the pattern - it has () -> Void listeners, not (Void) -> Void
```
HayesGordon, line 36: For some classes we won't be able to make use of this abstraction. But I think that is okay, we'll have 80% shared code, and things like Triggers and Lists will need to be custom.

PR #64, opened 2025-12-03
feat: Reanimated shared value → Rive property binding

mfazekas:
## Summary
- Added `playIfNeeded()` to make sure rive graphics is not in steady state after update
- Adds example demonstrating driving Rive properties from Reanimated shared values
- Uses `NitroModules.box()` to share `ViewModelNumberProperty` with worklets
- Updates happen synchronously on UI thread - no async bridging

## How it works
```typescript
const boxedProperty = useMemo(() => {
  const posYProperty = instance.numberProperty('posY');
  return NitroModules.box(posYProperty);
}, [instance]);

useAnimatedReaction(
  () => progress.value,
  (value) => {
    'worklet';
    const property = boxedProperty.unbox() as ViewModelNumberProperty;
    property.value = value;  // Direct setter on UI thread!
   }
);
```

## Test plan
- Run the "Reanimated Shared Value" example
- Blue circle (React Native) and red circle (Rive) should animate in sync

In slow motion there is up to 1 frame difference:

### DIFF
--- a/example/src/pages/SharedValueListenerExample.tsx
+++ b/example/src/pages/SharedValueListenerExample.tsx
@@ -125,7 +125,6 @@ function AnimatedRiveExample({
   }));
 
   const animateTo800 = useCallback(() => {
-    progress.value = 0;
     progress.value = withSpring(800, {
       damping: 8,
       stiffness: 80,


nitrogen/generated/android/c++/JFunc_void.hpp
```diff
@@ -11,6 +11,7 @@
 #include <functional>
 
 #include <functional>
+#include <NitroModules/JNICallable.hpp>
```
HayesGordon, line 14: Is this nitrogen change just something missing from a previous PR?

HayesGordon (approved):
Nice!

nitrogen/generated/android/c++/JFunc_void.hpp
mfazekas: Yes I tried to update nitrogen, but now reverted

PR #71, opened 2025-12-09
feat: add data binding lists support

### LINKED ISSUE #11: Add data binding lists support
See [Data binding lists](https://rive.app/docs/runtimes/data-binding#lists)

mfazekas:
## Summary
 Adds `ViewModelListProperty` for dynamic list management at runtime, aligned with rive-react API.

 ### API
  - `length` - get list size
  - `getInstanceAt(index)` - get ViewModelInstance at index
  - `addInstance(instance)` - add to end of list
  - `addInstanceAt(instance, index)` - insert at position
  - `removeInstance(instance)` - remove by reference
  - `removeInstanceAt(index)` - remove at index
  - `swap(index1, index2)` - swap items
  - `addListener(callback)` - observe changes

 ### Example
  ```tsx
  const listProperty = instance.listProperty('items');

  // Add new item
  const newItem = viewModel.createInstance();
  listProperty.addInstance(newItem);

  // Remove item
  listProperty.removeInstanceAt(0);

  // Swap items
  listProperty.swap(0, 1);
```

# Test plan

  - Added example page with list.riv test file
  - iOS implementation
  - Android implementation
  
Closes #11

### DIFF
--- a/src/specs/ViewModel.nitro.ts
+++ b/src/specs/ViewModel.nitro.ts
@@ -149,6 +149,6 @@ export interface ViewModelListProperty
   removeInstanceAt(index: number): void;
   /** Swap the instances at the given indices, returns true if successful */
   swap(index1: number, index2: number): boolean;
-  /** Add a listener to be notified when the list changes */
-  addListener(onChanged: () => void): void;
+  /** Add a listener to be notified when the list changes. Returns a function to remove the listener. */
+  addListener(onChanged: () => void): () => void;
 }
--- a/android/src/main/java/com/margelo/nitro/rive/HybridViewModelListProperty.kt
+++ b/android/src/main/java/com/margelo/nitro/rive/HybridViewModelListProperty.kt
@@ -56,8 +56,9 @@ class HybridViewModelListProperty(private val listProperty: ViewModelListPropert
     return true
   }
 
-  override fun addListener(onChanged: () -> Unit) {
-    listeners.add { _: Unit -> onChanged() }
+  override fun addListener(onChanged: () -> Unit): () -> Unit {
+    val remover = addListenerInternal { _ -> onChanged() }
     ensureValueListenerJob(listProperty.valueFlow.map { })
+    return remover
   }
 }
--- a/ios/HybridViewModelListProperty.swift
+++ b/ios/HybridViewModelListProperty.swift
@@ -56,7 +56,7 @@ class HybridViewModelListProperty: HybridViewModelListPropertySpec, ValuedProper
     return true
   }
 
-  func addListener(onChanged: @escaping () -> Void) throws {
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: src/specs/ViewModel.nitro.ts [hunk 1]
```diff
@@ -124,3 +127,28 @@ export interface ViewModelImageProperty
   /** Add a listener to the view model image property */
   addListener(onChanged: () => void): void;
 }
+
+/**
+ * A list property that contains a dynamic collection of {@link ViewModelInstance} objects.
+ * @see {@link https://rive.app/docs/runtimes/data-binding#lists Rive Data Binding Lists}
+ */
+export interface ViewModelListProperty
```
HayesGordon, line 135: Two things to consider:
- We don't HAVE to name these similar to android/ios. We can use the naming convention that matches what is expected when working with lists in RN.
- We could add additional methods/overloads if those are commonly used in RN. For example in Dart I added `firstOrNull` instead of throwing, and overloads for `operator []` and `operator[]=`. Maybe something like this possible here to

```
  @override
  ViewModelInstance operator [](int index) => instanceAt(index);

  @override
  void operator []=(
    int index,
    covariant FFIRiveViewModelInstanceRuntime value,
  ) =>
      insert(index, value);
```

android/src/main/java/com/margelo/nitro/rive/HybridViewModelInstance.kt
```diff
@@ -11,66 +11,45 @@ class HybridViewModelInstance(val viewModelInstance: ViewModelInstance) : Hybrid
   override val instanceName: String
     get() = viewModelInstance.name
 
-  override fun numberProperty(path: String): HybridViewModelNumberPropertySpec? {
-    try {
-        val numberProper = viewModelInstance.getNumberProperty(path)
-        return HybridViewModelNumberProperty(numberProper)
+  // Returns null if ViewModelException is thrown for iOS parity
```
HayesGordon, line 14: Good catch!

HayesGordon: Oh! This is what we did from the start. Nice abstration then!

src/specs/ViewModel.nitro.ts
mfazekas: I ended up using the `rive-react` naming for consistency. 

```tsx
  - getInstanceAt(index)
  - addInstance(instance)
  - addInstanceAt(instance, index) → boolean
  - removeInstance(instance)
  - removeInstanceAt(index)
  - swap(index1, index2) → boolean
  - length
  ```
  
  I think that will work for now, and we can add convince/shorter names later, if needed. 
  
  We could override operator [] and [] = in js with proxy but it's PITA, and not usually done.

PR #73, opened 2025-12-09
ci: add bump version workflow

mfazekas:


.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,43 @@
...
+  contents: write
+
+jobs:
+  bump-version:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
HayesGordon: This might need to be `secrets.PAT_GITHUB`

PR #78, opened 2025-12-10
fix(ios): validate artboard and state machine names before creating RiveViewModel

mfazekas:
## Summary
- Validates artboard and state machine names before creating `RiveViewModel` on iOS
- This prevents crashes when invalid names are provided, as the upstream Rive iOS SDK uses `try!` internally ([source](https://github.com/rive-app/rive-ios/blob/main/Source/RiveViewModel.swift#L156)) which crashes instead of propagating errors
- Errors are now properly bubbled up via the `onError` callback

## Reproducer
Pass an invalid `artboardName` or `stateMachineName` prop to `RiveView` - without this fix, the app crashes:

```tsx
<RiveView
  style={styles.rive}
  autoPlay={true}
  fit={Fit.Contain}
  file={riveFile}
  artboardName="InvalidArtboardNameThatDoesNotExist"
  onError={(err) => console.error('RiveView onError:', err)}
/>
```

### DIFF
--- a/ios/RiveReactNativeView.swift
+++ b/ios/RiveReactNativeView.swift
@@ -58,6 +58,8 @@ class RiveReactNativeView: UIView, RiveStateMachineDelegate {
   func configure(_ config: ViewConfiguration, dataBindingChanged: Bool = false, reload: Bool = false, initialUpdate: Bool = false) throws {
     if reload {
       cleanup()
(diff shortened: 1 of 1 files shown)


HayesGordon (approved):
LGTM! I wonder if we should do the same check on Android, just for consistency.

PR #82, opened 2025-12-11
docs: add rive native badges and update status

HayesGordon:
Quite like this! It shows the native iOS and Android versions that is used with the package!


mfazekas (approved):
Cool!

PR #87, opened 2025-12-15
feat: add QuickStart example for data binding

mfazekas:
Port of the QuickStart example from rive-react-native to demonstrate basic data binding with useRiveNumber and useRiveTrigger hooks.

### DIFF
--- a/example/src/pages/QuickStart.tsx
+++ b/example/src/pages/QuickStart.tsx
@@ -7,7 +7,7 @@
 */
 
 import { useEffect, useMemo } from 'react';
-import { Button, ScrollView, StyleSheet } from 'react-native';
+import { Button, View, StyleSheet } from 'react-native';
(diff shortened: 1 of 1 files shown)


example/src/pages/QuickStart.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  return (
+    <SafeAreaView style={styles.safeAreaViewContainer}>
+      <ScrollView contentContainerStyle={styles.container}>
+        {riveFile && (
+          <RiveView
+            hybridRef={setHybridRef}
+            file={riveFile}
+            fit={Fit.Layout}
+            style={styles.animation}
```
HayesGordon: let's always name the style `rive` or `graphic`. Moving away from just referring to Rive as an `animation`

PR #95, opened 2025-12-22
feat: add data binding artboards support

mfazekas:
## Summary
- Add support for swapping artboards at runtime via data binding properties
- Enables composing complex scenes from artboards in different Rive files

## New APIs
- `RiveFile.getBindableArtboard(name)` - create bindable artboard reference
- `RiveFile.artboardNames` / `artboardCount` - enumerate artboards  
- `ViewModelInstance.artboardProperty(path)` - get artboard property
- `ViewModelArtboardProperty.set(artboard)` - swap artboard at runtime

## Example
```tsx
const { riveFile: mainFile } = useRiveFile(require('./scene.riv'));
const { riveFile: assetsFile } = useRiveFile(require('./characters.riv'));

const instance = mainFile?.defaultArtboardViewModel()?.createDefaultInstance();
const artboardProp = instance?.artboardProperty('CharacterArtboard');

// Swap to a different character
const character = assetsFile?.getBindableArtboard('Character 1');
artboardProp?.set(character);
```

https://github.com/user-attachments/assets/ce44488d-c659-4c5c-8cd8-e40e268c24d6

## References
- [Rive Data Binding Docs](https://rive.app/docs/runtimes/data-binding#artboards)
- [rive-react example](https://codesandbox.io/p/sandbox/rive-react-data-binding-artboards-kmvzh8)


android/src/main/java/com/margelo/nitro/rive/HybridViewModelArtboardProperty.kt
```diff
@@ -0,0 +1,25 @@
...
+@DoNotStrip
+class HybridViewModelArtboardProperty(private val property: ViewModelArtboardProperty) :
+  HybridViewModelArtboardPropertySpec() {
+
+  override fun set(artboard: HybridBindableArtboardSpec?) {
+    val bindable = (artboard as? HybridBindableArtboard)?.bindableArtboard
+    property.set(bindable)
+  }
+
+  // Android SDK doesn't support listeners on artboard properties
```
HayesGordon: @ErikUggeldahl and @dskuza looks like another inconsistency between Android and iOS.

@mfazekas lets not expose listening on artboard properties for the time being until both platforms have it. Otherwise people might test that it works on iOS and expect the same on Android.

example/src/pages/DataBindingArtboardsExample.tsx
```diff
@@ -0,0 +1,280 @@
...
+  Fit,
+  RiveView,
+  useRiveFile,
+  type RiveFile,
+  type BindableArtboard,
+} from '@rive-app/react-native';
+import { type Metadata } from '../helpers/metadata';
+
+/**
+ * Data Binding Artboards Example
```
HayesGordon, line 19: Really good explanation. We should also update the other examples to link to the marketplace!

HayesGordon (changes requested):
Looks great! But let's remove the ability to add a listener for the time being until Android supports it.

HayesGordon (approved):
Thanks! looks great

PR #106, opened 2026-01-08
feat: add in-app test runner

mfazekas:
## Summary
- Add in-app test runner using react-native-harness for assertions
- Add Tests page to example app for running tests on device
- Initial tests for `viewModel()` and `replaceViewModel()` from #96

## Test plan
- Run example app and navigate to "Tests" page
- Tap "Run All Tests" to verify both tests pass

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,6 +209,7 @@ jobs:
           yarn turbo run build:ios --cache-dir="${{ env.TURBO_CACHE_DIR }}"
 
   test-harness-ios:
+    if: false # TODO: re-enable once harness timeout issues are resolved
     runs-on: macos-latest
     env:
       XCODE_VERSION: 16.4


.github/workflows/ci.yml
```diff
@@ -207,3 +207,62 @@ jobs:
           TURBO_FORCE: ${{ steps.turbo-cache-ios.outputs.cache-hit != 'true' }}
         run: |
           yarn turbo run build:ios --cache-dir="${{ env.TURBO_CACHE_DIR }}"
+
+  test-harness-ios:
+    if: false # TODO: re-enable once harness timeout issues are resolved
+    runs-on: macos-latest
```
HayesGordon, line 213: We should update our runners to use Cirrus:
```suggestion
    runs-on: ghcr.io/cirruslabs/macos-runner:sonoma
```

PR #123, opened 2026-01-14
chore(tests): resolve duplicate react-native instances in harness tests

mfazekas:
## Summary
- Fix Metro config to resolve duplicate react-native instances causing `require()` to fail in harness tests
- Extract `withSingleReactNative` helper to `example/metro.helpers.js` for reuse
- Simplify test architecture: tests defined in `__tests__/*.harness.ts` with `describe`/`it`
- TestsPage uses `require.context` + `getTestCollector().collect()` to run same tests in-app

## Test plan
- [x] `yarn typecheck` passes
- [x] `yarn test:harness:ios` passes (4 tests)
- [x] In-app TestsPage loads and runs tests

### DIFF
--- a/example/metro.helpers.js
+++ b/example/metro.helpers.js
@@ -12,6 +12,10 @@ const path = require('path');
 function withSingleReactNative(config, projectDir) {
   const rnPath = path.join(projectDir, 'node_modules/react-native/index.js');
   const originalResolveRequest = config.resolver.resolveRequest;
+  const defaultResolve = (context, moduleName, platform) =>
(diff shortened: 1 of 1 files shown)


example/babel.config.js
```diff
@@ -8,6 +8,7 @@ module.exports = getConfig(
   {
     presets: ['module:@react-native/babel-preset'],
     plugins: [
+      '@babel/plugin-transform-class-static-block',
```
HayesGordon, line 11: I ran into this locally as well.

PR #128, opened 2026-01-20
test(harness): add useRiveNumber hook test

mfazekas:
## Summary
- Add harness test for `useRiveNumber` hook
- Fix duplicate React instances in metro config that prevented hooks from working in harness tests

## Root Cause
The library's hooks were importing React from a different location than the harness, causing "Cannot read property 'useMemo' of null" errors. Fixed by adding `react` to the module deduplication list in `metro.helpers.js`.


HayesGordon (approved):
LGTM!

PR #129, opened 2026-01-20
feat: useViewModelInstance use name and artboardName when getting vmi from a file

mfazekas:
## Summary

Fixes #124 - `useViewModelInstance` hook now correctly handles the `name` parameter:

- `name` consistently means **ViewModel instance name** (not ViewModel class name) for both `RiveFile` and `ViewModel` sources
- New `artboardName` param for `RiveFile` source to specify which artboard's ViewModel to use
- TypeScript overloads enforce param availability by source type (e.g., `artboardName` only available for `RiveFile`)

## Test plan

- [x] Unit tests pass (`yarn test`)
- [x] TypeScript compiles (`yarn typecheck`)
- [x] Lint passes (`yarn lint`)
- [ ] Harness tests on device

### DIFF
--- a/src/hooks/useViewModelInstance.ts
+++ b/src/hooks/useViewModelInstance.ts
@@ -3,16 +3,9 @@ import type { ViewModel, ViewModelInstance } from '../specs/ViewModel.nitro';
 import type { RiveFile } from '../specs/RiveFile.nitro';
 import type { RiveViewRef } from '../index';
 import { callDispose } from '../core/callDispose';
+import { ArtboardByName } from '../specs/ArtboardBy';
 
-export interface UseViewModelInstanceParams {
-  /**
-   * Get a specifically named instance from the ViewModel.
-   */
-  name?: string;
-  /**
-   * Create a new (blank) instance from the ViewModel.
-   */
-  useNew?: boolean;
+interface UseViewModelInstanceBaseParams {
   /**
    * If true, throws an error when the instance cannot be obtained.
    * This is useful with Error Boundaries and ensures TypeScript knows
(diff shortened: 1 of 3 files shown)


src/hooks/useViewModelInstance.ts
```diff
@@ -27,27 +20,57 @@ export interface UseViewModelInstanceParams {
   onInit?: (instance: ViewModelInstance) => void;
 }
 
+export interface UseViewModelInstanceFileParams
+  extends UseViewModelInstanceBaseParams {
+  /**
+   * The name of the artboard to get the ViewModel from.
+   * If not provided, uses the default artboard.
+   */
+  artboardName?: string;
```
HayesGordon, line 29: I like this, but I'm worried that we now have many ways to do the same thing. You could get the correct vmi by doing:
```
const viewModel = riveFile?.viewModelByName('My View Model');
const namedInstance = useViewModelInstance(viewModel, { name: 'My Instance' });
```
I'm worried that if we expose this convenience, we bury this usage. Which is important when you're working with lists.

But I see the convenience of this. Ideally, with the future API where you will have the Artboard and State Machine configured before creating the view, this additional parameter won't be needed as the artboard will already be known by then.

HayesGordon (approved):
LGTM! I left one comment, but this is something we will address in the ideal API when we adopt the new iOS and Android runtimes.

src/hooks/useViewModelInstance.ts
mfazekas: one option is extending the options to useViewModelInstance(file, { options })

so there could be `artboardName`, `viewModelName` and `instanceName`.

In this PR it's `arboardName` and `name`  (instead of `instanceName`). But `instanceName` would be better - more explicit. 


```tsx
const viewModel = riveFile?.viewModelByName('My View Model');
const namedInstance = useViewModelInstance(viewModel, { name: 'My Instance' });
```

Yep that's works too. The only thing is that since `const viewModel = riveFile?.viewModelByName('My View Model');` is not a hook there is a decision the user has to make if he wants to memoize it or not. Probably doesn't matter as viewModelByName is very fast I'd assume.


This would be the equivalent:
```
const namedInstance = useViewModelInstance(viewModel, { modelName: 'My View Model', instanceName: 'My Instance' });
```

PR #147, opened 2026-02-18
ci: add Android harness test job

mfazekas:
Adds Android harness test CI job mirroring the existing iOS harness setup. Also fixes two Android test failures exposed by running the harness.

- Android harness CI job with emulator (API 35, x86_64), 3 retries, 5-min per-attempt timeout
- Timeout guards: 60 min for iOS, 30 min for Android
- Fix colorProperty Double→Int overflow on Android (`toLong().toInt()`)
- Enum test tolerance for Android legacy SDK accepting invalid values

### DIFF
--- a/example/__tests__/rive.harness.ts
+++ b/example/__tests__/rive.harness.ts
@@ -60,8 +60,6 @@ describe('ViewModel', () => {
 
     const vm1AfterReplace = instance?.viewModel('vm1');
     const vm1NameProp = vm1AfterReplace?.stringProperty('name');
-    // Android experimental backend doesn't support replaceViewModel yet (no-op)
-    const replaceVal = vm1NameProp?.value;
-    expect(replaceVal === testValue || replaceVal === 'name1').toBe(true);
+    expect(vm1NameProp?.value).toBe(testValue);
   });
 });


android/src/main/java/com/margelo/nitro/rive/HybridViewModelColorProperty.kt
```diff
@@ -12,7 +12,7 @@ class HybridViewModelColorProperty(private val viewModelColor: ViewModelColorPro
   override var value: Double
     get() = viewModelColor.value.toDouble()
     set(value) {
-      viewModelColor.value = value.toInt()
+      viewModelColor.value = value.toLong().toInt()
```
HayesGordon, line 15: Good catch!

example/__tests__/viewmodel-properties.harness.ts
```diff
@@ -81,7 +81,10 @@ describe('ViewModel Properties', () => {
     expect(enumProperty.value).toBe('cat');
 
     enumProperty.value = 'snakeLizard';
-    expect(enumProperty.value).toBe('cat');
+    // Most backends reject invalid enum values; the value should revert to 'cat'
+    // Android legacy SDK accepts them (reads back 'snakeLizard')
+    const val = enumProperty.value;
+    expect(val === 'cat' || val === 'snakeLizard').toBe(true);
```
HayesGordon, line 87: If it's legacy that is doing it wrong then that's okay.

PR #152, opened 2026-02-20
feat: setting fallback fonts

mfazekas:
Adds `RiveFonts` — font fallback API for react-native. Two-phase design: load fonts into opaque handles, then configure a weight-based fallback map. System font fallback is opt-in via `RiveFonts.systemFallback()`.

```ts
const regular = await RiveFonts.loadFont({ uri: 'https://…/font.ttf' });
const bold = await RiveFonts.loadFont('kanit_regular.ttf');
const cjk = await RiveFonts.loadFont({ name: 'Noto Sans CJK' });

// Custom fonts only — no system fallback
await RiveFonts.setFallbackFonts({
  default: [regular, cjk],
  700: [bold, cjk],
});

// Custom fonts + system fallback (explicit opt-in)
await RiveFonts.setFallbackFonts({
  default: [regular, cjk, RiveFonts.systemFallback()],
});

// Reset
await RiveFonts.clearFallbackFonts();
```

### How to test

Open the **Font Fallback** exerciser in the example app (`Exercisers > Font Fallback`).

**1. Custom fonts only** — select fonts, mount without "System Fallback" toggle — unsupported scripts show tofu.

**2. Custom fonts + system fallback** — select fonts, toggle "System Fallback" on, mount — unsupported scripts render via platform default.

**3. Font ordering** — toggles show selection order (1, 2, 3…) since fallback priority follows insertion order.

| iOS | Android |
|-----|---------|
| image |image |

| iOS | Android |
|-----|---------|
| image |  image |


src/specs/Rive.nitro.ts
```diff
@@ -1,6 +0,0 @@
-import type { HybridObject } from 'react-native-nitro-modules';
-
-export interface Rive
```
HayesGordon, line 3: Ah this lasted long :)

I figured eventually this would get extended in some way

ios/HybridRiveFontConfig.swift
```diff
@@ -0,0 +1,114 @@
...
+    guard let font = UIFont(name: fontName, size: UIFont.systemFontSize) else {
+      throw RuntimeError.error(withMessage: "Failed to create UIFont for: \(fontName)")
+    }
+    return font
+  }
+
+  private static func updateFallbackFonts() {
+    var providers: [RiveFallbackFontProvider] = customFonts
+    providers.append(RiveFallbackFontDescriptor())
+    RiveFont.fallbackFonts = providers
```
dskuza: I'll mention here that there are two ways of setting fallback fonts in the Apple runtime: 

1. An array of fonts that will be applied to any weighted character
2. A callback that allows you to set various arrays of fonts based on what the missing character's weight is

I don't know if there's an Android complement, but I do know that this is an API some Apple runtime consumers use. If there is an Android complement, it might be nice to use that instead. Or, supply both options.

Thoughts, @HayesGordon?

ios/HybridRiveFontConfig.swift
```diff
@@ -0,0 +1,114 @@
+import CoreText
+import NitroModules
+import RiveRuntime
+
+class HybridRiveFontConfig: HybridRiveFontConfigSpec {
+  private static var systemFallbackEnabled = false
+  private static var customFonts: [UIFont] = []
+
+  func enableSystemFontFallback() throws -> Promise<Void> {
```
dskuza: On iOS, fallback fonts are always there (there are defaults if you don't provide one).

ios/HybridRiveFontConfig.swift
mfazekas: @dskuza thanks much reworked to allow weight specific override on both platforms

ios/HybridRiveFontConfig.swift
mfazekas: @dskuza thanks much, removed

android/src/main/java/com/margelo/nitro/rive/HybridRiveFontConfig.kt
```diff
@@ -0,0 +1,131 @@
...
+          val fonts = synchronized(fontsByWeight) {
+            fontsByWeight[requestedWeight] ?: fontsByWeight[0] ?: emptyList()
+          }
+          val result = fonts.toMutableList()
+          val systemFonts = FontHelper.getFallbackFonts(Fonts.FontOpts(weight = weight))
+          for (font in systemFonts) {
+            try {
+              FontHelper.getFontBytes(font)?.let { result.add(it) }
+            } catch (e: OutOfMemoryError) {
+              Log.w(TAG, "Skipped system font due to memory pressure: ${e.message}")
```
HayesGordon: Slight concern that we're burrying this catch

ios/HybridRiveFontConfig.swift
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  func applyFallbackFonts() throws -> Promise<Void> {
+    return Promise.async {
+      _ = RiveFont.self
+      RiveFont.fallbackFontsCallback = { weight in
+        let requestedWeight = Int(weight.rawWeight)
+        let fonts = Self.fontsByWeight[requestedWeight] ?? Self.fontsByWeight[0] ?? []
+        var providers: [RiveFallbackFontProvider] = fonts
+        providers.append(RiveFallbackFontDescriptor())
```
dskuza: Is this the way of "guaranteeing" there's at least one fallback font in use (if all else fail)? We don't do this on iOS (there's only a default if you don't provide an explicit list). Not saying this is wrong, just something to call out.

mfazekas: I've changed things around so system fallback have to be added manually we don't auto append it. it makes testing much easier. It's an advanced api, so I think it's better to be let the user decide it

mfazekas:
@edgarrii sorry I'm not sure what's going on, on the video. If I try to select serif that throws for me. Maybe you have a `serf` font installed that causes the issue?!

image

PR #153, opened 2026-02-20
feat: add RiveRuntime API for Android init failure handling

mfazekas:
## Summary

Moves `Rive.init()` out of `ContentProvider` (which crashes before user code can run) into `RivePackage.createViewManagers()` with try/catch. Exposes `RiveRuntime.initialize()` and `RiveRuntime.getStatus()` to JS so apps can detect and recover from native lib loading failures (ABI mismatch, missing .so, etc).

Refs: [rive-android#323](https://github.com/rive-app/rive-android/issues/323), [rive-android#341](https://github.com/rive-app/rive-android/issues/341)

Closes: #151

## API

```ts
import { RiveRuntime } from '@rive-app/react-native';

const { isInitialized, error } = RiveRuntime.getStatus();
await RiveRuntime.initialize(); // manual init
```

To skip auto-init, set `Rive_RiveRuntimeAndroidSkipSetup=true` in `gradle.properties`.

## Test plan

Tested with `Rive_RiveRuntimeAndroidExcludeNativeLibs=true` in our example. The code showed this:

image

image

### DIFF
--- a/ios/HybridRiveRuntime.swift
+++ b/ios/HybridRiveRuntime.swift
@@ -6,6 +6,6 @@ final class HybridRiveRuntime: HybridRiveRuntimeSpec {
   var initError: String? { nil }
 
   func initialize() throws -> Promise<Void> {
-    return .resolved(with: ())
+    return .resolved()
   }
 }


example/android/app/build.gradle
```diff
@@ -85,7 +85,20 @@ android {
         versionCode 1
         versionName "1.0"
         multiDexEnabled true
+
     }
+
+    // Exclude Rive native libs to emulate a native lib loading failure.
```
HayesGordon, line 91: Nice!

PR #213, opened 2026-04-14
fix: survive fast refresh without disposing native objects

### LINKED ISSUE #205: Fast refresh breaks Rive
### Description

This is only a development annoyance, but I think it's worth raising that fast refresh breaks the Rive animations:
```
 ERROR  [Error: Cannot get hybrid property `HybridViewModelInstanceSpec.instanceName` - `this`'s `NativeState` is `null`, did you accidentally call `dispose()` on this object?]
 ERROR  [Error: Should not already be working.]
```

### Provide a Repro

https://github.com/Nezz/expo-repro/tree/repro/RiveTransform

Repro steps:
1. `bun android` or `bun ios`
2. Add a line break to `RiveBorderButton`

#### Source `.riv`/`.rev` file

Included in the repo

### Expected behavior

Fast refresh should not break the app. Sometimes only animations break, other times the app has to be fully reloaded.

### Device & Versions (please complete the following information)
- Device: Samsung Galaxy Fold 3, API 36 emulator, iOS simulator
- OS: Android 15, iOS 26.4
- React Native Version: 0.83.4
- Expo Version (if applicable): 55

mfazekas:
## Summary

Cherry-pick of #212 onto main. Fixes #205.

Fast refresh (HMR) re-runs all effect cleanups then all effects. Hooks that disposed native HybridObjects in effect cleanup would destroy the backing native state, but `useMemo` caches still returned the same (now dead) JS references — causing `NativeState is null` crashes.

Introduces `useDisposableMemo` — a `useMemo` replacement for disposable native resources:
- Value created synchronously during render (available on first render)
- Old value disposed in render phase when deps change  
- **Prod**: synchronous dispose on unmount (zero overhead)
- **Dev**: deferred dispose via `setTimeout(0)`, cancelled if fast refresh / Strict Mode re-mounts the component

Migrates `useRiveProperty`, `useRiveList`, and `useViewModelInstance`.

## Test plan

- [x] 9 unit tests for useDisposableMemo
- [x] All existing tests pass
- [x] Tested on iOS with Data Binding exerciser — fast refresh survives

### DIFF
--- a/src/hooks/__tests__/useDisposableMemo.test.ts
+++ b/src/hooks/__tests__/useDisposableMemo.test.ts
@@ -149,11 +149,9 @@ describe('useDisposableMemo', () => {
 
     const { result, rerender } = renderHook(
       (props: { dep: string }) =>
-        useDisposableMemo(
-          () => undefined as Disposable | undefined,
-          cleanup,
-          [props.dep]
-        ),
+        useDisposableMemo(() => undefined as Disposable | undefined, cleanup, [
+          props.dep,
+        ]),
       { initialProps: { dep: 'A' } }
     );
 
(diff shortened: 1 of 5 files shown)


HayesGordon (approved):
LGTM!

PR #216, opened 2026-04-16
fix: bump rive-ios to 6.18.2 and rive-android to 11.4.0

mfazekas:
Bump native Rive SDKs: rive-ios 6.18.0 → 6.18.2, rive-android 11.3.1 → 11.4.0.

Note: rive-ios 6.18.3 is tagged on GitHub but not yet published to CocoaPods, so using 6.18.2.

### rive-ios 6.18.1–6.18.2
- fix(renderer): gamma correction fix
- fix(runtime): databind state machine transition duration
- fix(runtime): nullptr check for ConditionComparison during transition
- fix(runtime): memory leak with listeners on scripted VM properties
- fix(editor/runtime): honor path effect shape for inner vector feathering
- feat: expose view model name from view model instance
- feat: image fit & alignment when parented by layout

### rive-android 11.3.2–11.4.0
- fix(android): keep RenderTarget alive for deferred draw callbacks
- fix(Vulkan): fix saturation blend mode on some devices
- fix(runtime): databind state machine transition duration
- fix(runtime): nullptr check for ConditionComparison during transition
- fix(runtime): memory leak with listeners on scripted VM properties
- feat: component based conditions support
- feat: single/multiline support in TextInput and improved scrolling
- feat: image fit & alignment when parented by layout

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -1754,7 +1754,7 @@ PODS:
     - React-logger (= 0.79.2)
     - React-perflogger (= 0.79.2)
     - React-utils (= 0.79.2)
-  - RiveRuntime (6.18.0)
+  - RiveRuntime (6.18.2)
   - RNCAsyncStorage (2.2.0):
     - DoubleConversion
     - glog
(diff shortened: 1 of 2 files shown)


HayesGordon (approved):
LGTM!

PR #222, opened 2026-04-17
fix(iOS): dispatch completion to main thread in handlePreloadedImage

### LINKED ISSUE #221: iOS crash: `dispatchPrecondition` failure in `ReferencedAssetLoader.releaseFile()` when using preloaded images via `RiveImages.loadFromURLAsync`
## What we're trying to do

We want to render a Rive file **only after its referenced assets have finished loading**, so the first frame the user sees is already complete — no asset pop-in or flicker.

We first tried the Asset Handler API. It works functionally, but assets are fetched/decoded **asynchronously after the Rive file has already rendered**: `useRiveFile` reports the file as loaded even though its referenced assets aren't ready yet, so the animation renders once without the assets and again once they arrive, causing a visible flicker.

We were then betting on `RiveImages.loadFromURLAsync` + React Suspense to solve this: preload the image, only then construct the `RiveFile` with the preloaded `RiveImage` passed in via `referencedAssets`, and render. But this path crashes on iOS (see below), so we haven't been able to confirm whether it would produce the flicker-free UX we want.

## Summary

On iOS, passing a preloaded `RiveImage` (e.g. from `RiveImages.loadFromURLAsync`) into `useRiveFile`'s `referencedAssets` crashes with `EXC_BREAKPOINT`:

```
BUG IN CLIENT OF LIBDISPATCH: Assertion failed: Block was expected to execute on queue
[com.apple.main-thread (0x1feeb5b80)] > com.apple.main-thread
```

## Environment

- Platform: iOS (reproduced on iOS Simulator, MacBook Pro M1 Pro, macOS 26.3.1)
- `@rive-app/react-native` **0.4.1 - 0.4.3** 
- React Native: **0.79.2 - 0.83.2** 
- `RiveRuntime` (iOS) 6.18.2
- New Architecture: enabled

## Reproduction

Run the example app's **Out-of-Band Assets (Suspense)** exerciser (`example/src/exercisers/OutOfBandAssetsWithSuspense.tsx`), which calls `RiveImages.loadFromURLAsync(...)` and passes the resulting `RiveImage` to `useRiveFile` under `referencedAssets`. The app crashes as soon as the `RiveFile` is created.

Passing the same asset via a URL/file `source` (letting the loader fetch and decode it) does **not** crash — only the preloaded-image path does.

## Crash trace (key frames)

```
Thread 3 Crashed:  Dispatch queue: com.apple.root.user-initiated-qos
0  libdispatch            _dispatch_assert_queue_fail
1  libswiftDispatch       _dispatchPreconditionTest(_:)
2  ReferencedAssetLoader.releaseFile()                         ReferencedAssetLoader.swift:26
3  closure #1 in closure #1 in createCustomLoader(...)         ReferencedAssetLoader.swift:159
4  ReferencedAssetLoader.handlePreloadedImage(...)             ReferencedAssetLoader.swift:89
5  ReferencedAssetLoader.loadAssetInternal(...)                ReferencedAssetLoader.swift:97
6  closure #1 in createCustomLoader(...)                       ReferencedAssetLoader.swift:156
7  RiveRuntime            rive::FileAssetLoaderAdapter::loadContents(...)
8  RiveRuntime            rive::File::import(...)
```

## Potential Fix

In `ios/ReferencedAssetLoader.swift`, `handlePreloadedImage` calls `completion()` synchronously on the Rive C++ runtime's background queue. That `completion` is `{ self?.releaseFile() }`, and `releaseFile()` asserts it must run on the main queue — hence the crash. Every other asset path already hops to main before calling `completion()` (`processAssetBytes` uses `DispatchQueue.main.async`; the `Task` branch uses `await MainActor.run`); only `handlePreloadedImage` is missing this hop.

Wrap its body in `DispatchQueue.main.async` to match:

```swift
private func handlePreloadedImage(
  _ image: any HybridRiveImageSpec, asset: RiveFileAsset, completion: @escaping () -> Void
) {
  guard let imageAsset = asset as? RiveImageAsset,
    let hybridImage = image as? HybridRiveImage
  else {
    DispatchQueue.main.async { completion() }
    return
  }

  DispatchQueue.main.async {
    imageAsset.renderImage(hybridImage.renderImage)
    completion()
  }
}
```

This also ensures `imageAsset.renderImage(...)` (a UIKit-adjacent mutation) runs on the main thread.

mfazekas:
Fixes #221

`handlePreloadedImage` was calling `completion()` synchronously on the Rive C++ runtime's background import queue, but `completion` triggers `releaseFile()` which asserts main thread via `dispatchPrecondition`. Every other asset path in `ReferencedAssetLoader` already dispatches to main before calling `completion()` — this was the only one missing it.

### Crash stack (reproduced locally)
```
_assertionFailure
ReferencedAssetLoader.handlePreloadedImage(_:asset:completion:)
ReferencedAssetLoader.loadAssetInternal(source:asset:factory:completion:)
closure #1 in ReferencedAssetLoader.createCustomLoader(...)
rive::File::import(...)  ← background queue
HybridRiveFileFactory.genericFrom(...)  ← DispatchQueue.global(.userInitiated)
```

Reproduced using the Out-of-Band Assets (Suspense) exerciser which calls `RiveImages.loadFromURLAsync` and passes the preloaded `RiveImage` via `referencedAssets`.

### DIFF
--- a/ios/ReferencedAssetLoader.swift
+++ b/ios/ReferencedAssetLoader.swift
@@ -81,12 +81,14 @@ final class ReferencedAssetLoader {
     guard let imageAsset = asset as? RiveImageAsset,
       let hybridImage = image as? HybridRiveImage
     else {
-      completion()
+      DispatchQueue.main.async { completion() }
       return
     }
 
-    imageAsset.renderImage(hybridImage.renderImage)
-    completion()
+    DispatchQueue.main.async {
+      imageAsset.renderImage(hybridImage.renderImage)
+      completion()
+    }
   }
 
   private func loadAssetInternal(


HayesGordon (approved):
LGTM!

PR #226, opened 2026-04-20
fix(Android): prevent data binding blink on mount

### LINKED ISSUE #203: Android animation enters wrong state on mount
### Description

When starting an animation on Android, it ends up in the wrong state initially and then fixes itself. This results in the animation "blinking" on mount. This issue does not reproduce on iOS.

This regressed when updating from 0.2.3 to 0.4.0.

Note that during the very first frame the layout is also broken and the animation is so large that it ends up getting clipped:
Image

### Provide a Repro

https://github.com/Nezz/expo-repro/tree/repro/RiveAndroid

#### Source `.riv`/`.rev` file

Included in the repo

### Expected behavior

The border animation should not play right away (which is already the case on iOS). Only when the button is tapped.

#### Screenshots

https://github.com/user-attachments/assets/05718591-96bc-444e-8208-acc05cd2c73d

### Device & Versions (please complete the following information)
- Device: Samsung Galaxy Fold 3, API 36 emulator
- OS: Android 15
- React Native Version: 0.83.4
- Expo Version (if applicable): 55

mfazekas:
Fixes #203

When data binding is active, the state machine was being advanced (`settleInitialState`) before the ViewModel was bound, causing a single frame with wrong default values (visible as a border animation flash on mount).

Fix: call `play(settleInitialState=false)` to create the SM without advancing it, then bind the ViewModel, so the first Choreographer frame advances with correct values already bound.

https://github.com/user-attachments/assets/placeholder-before

https://github.com/user-attachments/assets/placeholder-after

### DIFF
--- a/android/src/main/java/com/rive/RiveReactNativeView.kt
+++ b/android/src/main/java/com/rive/RiveReactNativeView.kt
@@ -107,20 +107,24 @@ class RiveReactNativeView(context: ThemedReactContext) : FrameLayout(context) {
       ?: resources.displayMetrics.density
 
     if (reload) {
-      val hasDataBinding = when (config.bindData) {
-        is BindData.None -> false
-        is BindData.Auto -> false
-        is BindData.Instance, is BindData.ByName -> true
-      }
+      val hasDataBinding = config.bindData !is BindData.None
       riveAnimationView?.setRiveFile(
         config.riveFile,
         artboardName = config.artboardName,
         stateMachineName = config.stateMachineName,
-        autoplay = config.autoPlay,
-        autoBind = hasDataBinding,
+        autoplay = false,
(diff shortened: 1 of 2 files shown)


mfazekas:
Closing — this is not a library bug. The blink happens when VM properties are set in `useEffect` (fires after render) instead of using the `onInit` callback on `useViewModelInstance`, which runs synchronously before the instance reaches `<RiveView>`.

Correct pattern:
```tsx
const { instance } = useViewModelInstance(riveFile, {
  onInit: (vmi) => {
    vmi.booleanProperty('isFocused')?.set(false);
  },
});
```

This ensures the VM instance has the correct values before the state machine settles on the first frame.

PR #238, opened 2026-04-30
feat: add async variants for ViewModelListProperty mutation methods

mfazekas:
Deprecate sync addInstance, addInstanceAt, removeInstance, removeInstanceAt, swap in favor of async versions that return Promise<void>. Ported from feat/rive-ios-experimental branch.


HayesGordon (approved):
LGTM!

PR #260, opened 2026-05-15
ci: debug Android npx resolution failure

mfazekas:
Debug PR to investigate why `npx` is not found (exit 127) during Gradle settings evaluation in the Android harness test jobs. Adds PATH, disk space, and node_modules diagnostics before the Gradle build step.


mfazekas:
Closing — the Android CI failure was transient. Debug output from the passing run confirmed npx, node, and disk space are all fine. Keeping the debug info in the PR for reference if it recurs.

PR #262, opened 2026-05-18
fix: separate trigger hook lifecycle to prevent missed events on Android

### LINKED ISSUE #230: Android: unstable callback references can make `useRiveTrigger` and `onEventReceived` miss events
### Summary

On Android, after updating `@rive-app/react-native` from `0.4.2` to `0.4.5`, we found a case where Rive trigger/event listeners can silently stop firing if the callback reference changes across React re-renders.

This does **not** reproduce on iOS with the same `.riv` files and the same JS logic.

A practical workaround on our side was to make callbacks stable with `useEvent` / `useRef + useCallback([])`.

This looks related to listener/property re-subscription timing and may be adjacent to #156, but the repro here does **not** depend on `autoPlay={false}`.

### Environment

- `@rive-app/react-native`: `0.4.5`
- React Native: `0.84.1`
- Platform: Android only
- Previous working version: `0.4.2`

### Repro pattern

We use ViewModel triggers and event listeners from React components that can re-render multiple times while the screen is mounted.

Problematic shape:

```tsx
const { instance: riveInstance } = useViewModelInstance(riveFile, {
  instanceName: theme.isDarkTheme ? 'dark' : 'light',
  viewModelName: 'Theme',
});

useRiveTrigger('animationFinish', riveInstance, {
  onTrigger: () => {
    onAnimationEnd();
    wasAnimationPlayed.current = true;
  },
});
```
and similarly:
```tsx
<RiveAnimation
  onEventReceived={(event) => {
    if (event.name === 'anim_finish') {
      showButton();
    }
  }}
/>
```
In our app, these components re-render several times while mounted. On Android, after the update to 0.4.5, the listener may stop receiving the expected trigger/event.

### Workaround
Making the callback reference stable fixes the issue for us:
```tsx
const handleAnimationFinish = useEvent(() => {
  onAnimationEnd();
  wasAnimationPlayed.current = true;
});

useRiveTrigger('animationFinish', riveInstance, {
  onTrigger: handleAnimationFinish,
});
```
### Expected behavior
Changing the JS callback reference between renders should not cause Android trigger/event delivery to become unreliable.

### Actual behavior
On Android, useRiveTrigger(..., { onTrigger }) and event listeners passed through component props can miss expected events/triggers when callback references change across renders.

### Notes
We noticed this after the hook/runtime changes introduced in the newer 0.4.x line. From reading the runtime code, it looks like listener/property lifecycle may now be more sensitive to option identity changes during re-render.

mfazekas:
Fixes #230

Unstable `onTrigger` callback references cause `useRiveTrigger` to silently stop receiving events on Android after re-renders.

`useRiveTrigger` now manages its own property lifecycle (like `useRiveList`) instead of going through `useRiveProperty`. The callback is stored in a ref so its identity doesn't affect native property disposal. `useRiveProperty` is simplified — the `onPropertyEventOverride` escape hatch is removed and the `options` parameter becomes a plain `getProperty` function.

<details>
<summary>Root cause and reproducer</summary>

### Root cause

`useDisposableMemo` disposes the old native property and creates a new one **during the render phase** (via ref mutation). When the callback identity changes (e.g. `onTrigger` without React Compiler memoization), `options` changes, which triggers this dispose/create cycle.

The problem: if a no-op state update (like `setError(null)` when error is already null) triggers a re-render that produces **identical JSX output**, React **bails out** — it runs the component function but skips the commit and effects. Since `useEffect` is skipped, the listener is never re-subscribed to the new property.

```
useDisposableMemo:  deps changed → DISPOSE old, CREATE new   ✅ runs (render phase)
useEffect:          deps changed → should re-subscribe...    ❌ skipped (bailout)
```

The property is orphaned — disposed old one is dead, new one has no listener. All subsequent trigger events are silently lost.

### Reproducer

```tsx
import { useState } from 'react';
import { View, Text } from 'react-native';
import {
  RiveView, useRiveFile, useRiveTrigger,
  useViewModelInstance, Fit,

export default function Issue230() {
  'use no memo'; // disable React Compiler → unstable callbacks

  const [triggerCount, setTriggerCount] = useState(0);
  const { riveFile } = useRiveFile(require('./rewards.riv'));
  const { instance } = useViewModelInstance(riveFile);

  // New reference every render — this is what triggers the bug
  const onTrigger = () => setTriggerCount((c) => c + 1);

  useRiveTrigger('Button/Pressed', instance, { onTrigger });

  return (
    <View>
      <Text>Triggers received: {triggerCount}</Text>
      {riveFile && instance && (
        <RiveView file={riveFile} fit={Fit.Contain} dataBind={instance}
          style={{ width: '100%', height: 200 }} />
      )}
    </View>
  );
}
// Tap the Rive button — first tap works, then triggers stop.
// On Android the trigger fires at the native level (chest opens)
// but the JS callback is never called after the bailout render.
```

</details>

### DIFF
--- a/example/__tests__/use-rive-trigger.harness.tsx
+++ b/example/__tests__/use-rive-trigger.harness.tsx
@@ -1,0 +1,198 @@
+import {
+  describe,
+  it,
+  expect,
+  render,
+  waitFor,
+  cleanup,
+} from 'react-native-harness';
+import { useEffect, useState } from 'react';
+import { Text, View } from 'react-native';
+import { RiveFileFactory, useRiveTrigger } from '@rive-app/react-native';
+import type { ViewModelInstance } from '@rive-app/react-native';
+
+const DATABINDING = require('../assets/rive/databinding.riv');
+
+function expectDefined<T>(value: T): asserts value is NonNullable<T> {
+  expect(value).toBeDefined();
+}
+
+async function loadGordonInstance() {
+  const file = await RiveFileFactory.fromSource(DATABINDING, undefined);
+  const vm = file.viewModelByName('Person');
+  expectDefined(vm);
+  const instance = vm.createInstanceByName('Gordon');
+  expectDefined(instance);
+  return { file, instance };
+}
+
+// ─── Test context types ────────────────────────────────────────────
+
+type TriggerContext = {
+  triggerCount: number;
+  triggerFn: (() => void) | null;
+  error: Error | null;
+  renderCount: number;
+};
+
+function createTriggerContext(): TriggerContext {
+  return { triggerCount: 0, triggerFn: null, error: null, renderCount: 0 };
+}
+
+// ─── Test component: stable callback ───────────────────────────────
+
+function StableTriggerComponent({
+  instance,
+  context,
+}: {
+  instance: ViewModelInstance;
+  context: TriggerContext;
+}) {
+  context.renderCount++;
+
+  const { trigger, error } = useRiveTrigger('jump', instance, {
+    onTrigger: () => {
+      context.triggerCount++;
+    },
+  });
+
+  useEffect(() => {
+    context.triggerFn = trigger;
+    context.error = error;
+  }, [context, trigger, error]);
+
+  return (
+    <View>
+      <Text>{context.triggerCount}</Text>
+    </View>
+  );
+}
+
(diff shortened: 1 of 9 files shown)


HayesGordon (approved):
LGTM!

PR #269, opened 2026-05-21
chore: add standardized github issue template to react native runtime

zplata:
Adding a "standardized" github issue template similar to what we did with android/ios/js/etc. so users can fill out a form when submitting issues.

Feel free to suggest other questions/context we should gather in the issue template


mfazekas (approved):

mfazekas:
@zplata thanks, looks nice to me!

PR #293, opened 2026-06-15
chore: remove Early Release banner, mark artboards and value props roadmap items complete

mfazekas:
Removes the Early Release warning banner and updates the roadmap to reflect that Data Binding - Artboards and Data Binding - Value props are both fully implemented.


HayesGordon (approved):
🙌

PR #301, opened 2026-06-29
fix(android): avoid NoSuchElementException in view-model property listener dispatch

mfazekas:
`onChanged()` copied the listener map with `listeners.values.toList()`. Kotlin's `toList()` has a `size == 1` fast-path that calls `iterator().next()` unguarded, so it throws `NoSuchElementException` when the sole listener is removed concurrently (dispatch runs on `Dispatchers.Default` while the JS thread mutates) — crashing the app (`test-harness-android` *"App bridge disconnected"*).

Iterate the `ConcurrentHashMap` directly instead — its weakly-consistent iterator is `hasNext()`-guarded and never throws.


HayesGordon (approved):
LGTM!

Issue #308, opened 2026-07-01
[ci-flake] android-harness-native-crash — off-thread data-binding SIGSEGV


mfazekas:
**Signature:** `android-harness-native-crash`
**Regex:** (`SIGSEGV` | `bridge disconnected` | `global reference table overflow`) in the harness **test** step
**Category:** product-bug
**First seen:** 2026-06-29 · **Last seen:** 2026-07-06

## Occurrence log
- 2026-06-29 — `test-harness-android` — intermittent "App bridge disconnected" = native heap-corruption SIGSEGV (reproduced locally ~1-in-10 runs)
- 2026-07-06 — production (Discord, Sentry DISCORD-ANDROID-3NFN) — SIGSEGV in `rive::DataBindContainer::updateDataBinds` via `cppAdvance`; fixed on their side by rive-android 11.6.2 (see comment below; mitigation #317)

## Hypothesis
**This is a real bug, not infra** — do not just re-run it. The Rive data-binding runtime is accessed off the main thread (via `Promise.async` / the JS thread) without synchronization, corrupting the native heap. #298 fixed only the Java listener-map `ConcurrentModificationException`, **not** the underlying off-thread native access. Related: #297 / #298.

## Reproduction
- Page: `example/src/reproducers/Issue297ThreadRace.tsx` (toggleable concurrent access).
- Run the Android harness under stress; reproduces ~1-in-10. Confirm on iOS under **TSan** (`rn-harness.config.mjs` sets `TSAN_OPTIONS halt_on_error=0`) — 25 races observed previously.

## Mitigation
- [ ] Serialize all Rive runtime access to a single thread (or add native locking) so data-binding reads/writes never run off-main concurrently.

## Closing criterion (guard)
Native access is made thread-safe **and** a `*.harness.tsx` stress guard reproduces the old crash on an unfixed build / passes on the fixed one → move to `ci-flake:resolved`.

---
_Tracked via [CI Flake Triage](../blob/main/docs/ci-flake-triage.md). This issue is intentionally **closed** — it is a ledger entry, not open work._

mfazekas:
- 2026-07-01 — \`test-harness-android\` (job failed, run cancelled at ~35m after 3/3 attempts) — https://github.com/rive-app/rive-nitro-react-native/actions/runs/28504019421 — PR #304 (\`feat/use-viewmodel-instance-async\`)

[...]

PR #312, opened 2026-07-01
fix(android): bump rive-android to 11.6.1 (fixes data-binding UAF crash)

mfazekas:
Bumps `runtimeVersions.android` 11.4.1 → 11.6.1 to fix the intermittent native `SIGSEGV` in `test-harness-android` (experimental backend).

Root cause is a single-threaded re-entrancy / iterator-invalidation **use-after-free** in `rive::DataBindContainer::updateDataBinds` (a bind that auto-binds a nested view model re-enters `addDataBind`/`removeDataBind` and invalidates the active iterator). Fixed upstream in rive-runtime [#12649](https://github.com/rive-app/rive-runtime/pull/12649), shipped in rive-android 11.6.1 (11.4.1/11.5.0/11.6.0 are vulnerable). The experimental backend compiles + builds clean against 11.6.1 (no API changes needed).

Resolves the `android-databinding-native-crash` flake — see #308 for the full investigation.

### DIFF
--- a/example/__tests__/reconfigure.harness.tsx
+++ b/example/__tests__/reconfigure.harness.tsx
@@ -14,6 +14,7 @@ import {
   Fit,
   type RiveFile,
   type RiveViewRef,
+  type ViewModelNumberProperty,
 } from '@rive-app/react-native';
 
 const BOUNCING_BALL = require('../assets/rive/bouncing_ball.riv');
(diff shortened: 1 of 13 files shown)


mfazekas:
Superseded: #304 was rebased onto feat/rive-ios-experimental, which already carries rive-android 11.6.1 (via #313), so this direct bump is now a no-op. Closing.

PR #315, opened 2026-07-03
fix(ios): use C++17 for fmt pod (FMT_USE_CONSTEVAL=0 from #314 does not build)

mfazekas:
Follow-up to #314. The merged fmt workaround (define `FMT_USE_CONSTEVAL=0`) **does not actually fix** the Xcode 26.4+ (Apple clang 21) consteval error: `fmt/base.h` re-`#define`s `FMT_USE_CONSTEVAL` unconditionally (it is not `#ifndef`-guarded), so it overrides the command-line define and the build still fails. Verified failing on Xcode 26.6.

Switches to compiling the `fmt` pod as **C++17**, which makes fmt's own `base.h` disable consteval (`FMT_CPLUSPLUS < 201709` → `FMT_USE_CONSTEVAL 0`). **Verified building on Xcode 26.6.** CI (Xcode 26.3) does not hit the consteval error, so this only affects local builds on 26.4+.

Same approach as #306 (the main-line fmt PR). Real fix is upstream in React Native 0.83+.

### DIFF
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -34,21 +34,18 @@ target 'RiveExample' do
       # :ccache_enabled => true
     )
 
-    # fmt `consteval` build error on Xcode 26.4+ (Apple clang 21): the compiler rejects
-    # fmt's compile-time FMT_STRING (basic_format_string's consteval ctor) as a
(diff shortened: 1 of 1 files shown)


mfazekas:
Superseded by #349

PR #316, opened 2026-07-06
fix(hooks): warn and no-op when a property setter is called after dispose

mfazekas:
A setter returned by `useRiveNumber`/`useRiveString`/etc. can be captured in a stale closure (async callback, timeout) and fire after its property was disposed by a deps change or unmount. The write then reaches the disposed hybrid and throws `Cannot set hybrid property ... NativeState is null` — fatal when uncaught in a release app.

The fix routes writes through the `liveRef` that `useDisposableMemo` already maintains (nulled at dispose time) — the same guard `useRiveTrigger` uses. A write after unmount warns and no-ops (with the path and likely cause) instead of throwing; a write captured before a deps change targets the hook's current property, matching `useRiveTrigger` semantics. `property` stays in the `useCallback` deps so the setter identity still changes with the property.

Tests: jest cases for the after-unmount no-op+warn and the deps-change redirect, plus an on-device harness test (`staleSetterWrite.test.tsx`) that fails on the unguarded hook — verified on the Android emulator on both backends.

### DIFF
--- a/example/src/__tests__/staleSetterWrite.test.tsx
+++ b/example/src/__tests__/staleSetterWrite.test.tsx
@@ -1,0 +1,78 @@
+import { it, expect, render, waitFor, cleanup } from 'react-native-harness';
+import { useEffect } from 'react';
+import { View } from 'react-native';
+import {
+  Fit,
+  RiveView,
+  RiveFileFactory,
+  useRiveNumber,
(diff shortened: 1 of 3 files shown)


HayesGordon (approved):
LGTM!

PR #318, opened 2026-07-06
fix(android): don't re-acquire a disposed Rive file on view re-attach

mfazekas:
Fixes a production crash: after a RiveView is dropped, `dispose()` releases the controller (which releases its `File`), but `rendererAttributes.resource` still caches a `ResourceRiveFile` pointing at the now-disposed file. If the Android view is later attached to a window again (e.g. Fabric view recycling), `RiveAnimationView.onAttachedToWindow` reloads that stale resource and throws `Cannot acquire a disposed object` in `RiveFileController.setFile` -> `NativeObject.acquire`.

Legacy backend: `dispose()` now nulls `rendererAttributes.resource` before releasing the controller (mirrors `RiveAnimationView.saveControllerState()`), and `onAttachedToWindow` drops a cached `ResourceRiveFile` whose file has no cpp object before calling super. The new (CommandQueue) backend has no cached-resource reload path; its `dispose()` now also nulls `riveWorker` so a re-attach can't create a surface on a released command queue.

Reproduced and verified with the included `Disposed File Re-attach` reproducer page. The re-attach itself is driven by `rive-debug-utils`, a new private workspace package with an Android-only Nitro hybrid object that captures the dropped Android view and re-attaches it the way view recycling would; it autolinks into all three example apps. The reproducer crashes with the exact production stack before the fix and passes after.


rive-debug-utils/android/src/main/java/com/margelo/nitro/rivedebugutils/HybridDebugUtils.kt
```diff
@@ -0,0 +1,82 @@
...
+
+import android.view.View
+import android.view.ViewGroup
+import androidx.annotation.Keep
+import com.facebook.proguard.annotations.DoNotStrip
+import com.margelo.nitro.NitroModules
+import com.margelo.nitro.core.Promise
+
+/**
+ * Test helper for the example apps: simulates what Fabric view recycling does
```
HayesGordon, line 11: Super cool that you can do this!

HayesGordon (approved):
LGTM! The testing is great

mfazekas:
merged from main into experimental branch

PR #324, opened 2026-07-07
Experiment: Expo SDK 57 example with precompiled RN XCFrameworks + CI build timing

mfazekas:
Adds an `expo57-example` workspace (Expo SDK 57 / RN 0.86) and makes it the default iOS build on CI, using Expo's precompiled React Native XCFrameworks (https://expo.dev/blog/faster-ios-builds-with-precompiled-xcframeworks) plus the EAS prebuilt third-party bucket (reanimated/worklets; unofficial endpoint with automatic source fallback).

1. `build-ios` now builds expo57-example with caching (exact-key short-circuit, Pods cache, incremental build-dir cache). Cold xcodebuild is ~8 min vs ~26 min median for the bare example; timing details in the comments below.
2. `test-harness-expo57-ios` runs the shared harness suite against the Expo app. This needed the `unstable_path` asset publicPath (out-of-root shared assets) and a yarn patch on `@react-native-harness/bundler-metro` — its entry-point swap didn't recognize Expo's server-root-relative/virtual-entry requests, so the real app booted instead of the harness host (worth upstreaming).
3. The bare RN 0.79 example keeps its coverage via the existing `test-harness-ios`/`-legacy` jobs, which build it themselves.
4. expo57-example drops the create-expo-app theme boilerplate; its workspace typecheck is clean.

Verified locally (app renders, Quick Start data binding works, all 4 harness suites pass) and on CI (all jobs green).


mfazekas:
First timing results from run 28862816990 (both jobs cold, same runner pool, macos-15 / Xcode 26.3):

| job | pod install | xcodebuild (Debug, sim) |
| --- | --- | --- |
| build-expo57-ios (precompiled RN core + Expo modules) | 115s | 1013s (16.9 min) |
| build-ios (bare example, RN core from source, same run) | 99s | 1057s (17.6 min) |

Historical cold `build-ios` runs on main ranged 26–29 min of xcodebuild (runs 28794160443, 28573482962, 27693800321), so there's substantial runner variance; the same-run pair above is the fairest comparison.

Reading: the expo57 example compiles a much larger dependency set from source (reanimated, worklets, screens, gesture-handler, router native bits — third-party precompiled pods are EAS-only) plus RNRive/Rive, and still comes in slightly under the bare example. The precompiled RN core + Expo modules roughly offset the entire extra Expo/third-party surface. A few more runs will tighten the numbers.

mfazekas:
Second `build-expo57-ios` sample (rerun of the same commit, attempt 2): pod install 118s, xcodebuild 1332s (22.2 min) — vs 16.9 min in attempt 1, so runner variance clearly affects this job too.

Standings across all cold samples: expo57 16.9 / 22.2 min (n=2, mean ~19.5) vs bare example 17.6–31.3 min (n=9, median ~27). Both expo57 samples sit below the bare-example median even though the expo app compiles a much larger third-party set from source.

mfazekas:
With the EAS bucket enabled (run 28877621406): pod install 115s, xcodebuild 642s (**10.7 min**) — down from 16.9/22.2 min without third-party prebuilds. CI log confirms RNReanimated and RNWorklets installed as prebuilt XCFrameworks; screens/safe-area-context/async-storage 404'd for this version tuple and built from source as expected.

Cold-build summary so far: bare example median ~27 min (n=9) → expo57 precompiled RN core ~17–22 min (n=2) → expo57 + EAS third-party bucket ~10.7 min (n=1).

mfazekas:
test-harness-expo57-ios passed on its first CI run (run 28919441701): the full shared harness suite (4 suites) runs green against the Expo 57 app.

Timings: expo prebuild 6s, pod install 93s (reanimated + worklets prebuilt from the EAS bucket), xcodebuild 8m05s, harness tests 4m18s — **21m13s total, uncached**. For scale: the bare test-harness-ios/-legacy jobs in the same run (cold, same runner pool) were still building/running past the 40-minute mark when this one finished.

Two fixes were needed to run the harness against an Expo app (details in commit b3957da): the unstable_path publicPath form for out-of-root shared assets, and a yarn patch on @react-native-harness/bundler-metro so the entry-point swap recognizes Expo's server-root-relative / virtual-entry requests — worth upstreaming to Callstack.

mfazekas:
Final matrix after consolidating all harness jobs onto expo57-example (run 28925999512, all cold — cache keys changed):

| job | app build | tests | total |
| --- | --- | --- | --- |
| test-harness-ios | 12m14s | 2m24s | 24m35s |
| test-harness-ios-legacy | 18m06s | 5m04s | 38m49s |
| test-harness-android | 12m06s | 3m18s | 19m25s |
| test-harness-android-legacy | 12m39s | 3m21s | 20m15s |

All four passed on the first harness attempt. Notes: Android cold builds are ~4.5 min slower than the bare example's (~7.5 min) — the added Expo surface costs even with prebuilt Expo modules — but the test phase is faster and stabler (3m vs 11m with retries), so totals are a wash, and warm-cache runs skip the build entirely. Android already gets prebuilt RN core (Maven, since 0.71) and prebuilt Expo modules (SDK 53+ default, 15/18 modules as Maven artifacts in our build); there is no Android equivalent of the EAS third-party bucket.

Issue #332, opened 2026-07-10
Android CPU usage


ahmedu007:
### Submission checklist

- [x] I have confirmed the issue is present in the latest version of `@rive-app/react-native`
- [x] I have searched the documentation and forums and could not find an answer
- [x] I have searched existing issues and this is not a duplicate

### Description

After installing Rive into the app, the CPU usage reported by [flashlight.dev](https://flashlight.dev/) on Android has almost doubled

<img width="1616" height="629" alt="Image" src="https://github.com/user-attachments/assets/cc6a76c7-87a3-4b99-a97a-3ec4428c2355" />

### Reproduction steps / code

```tsx
import { DataBindMode, Fit, RiveView, useRive, useRiveFile } from '@rive-app/react-native';
import { useEffect } from 'react';
import { Dimensions, View } from 'react-native';
import { useTheme } from 'styled-components/native';

import { Text } from '~/themes/components/text';
import { messages } from './intl';

export const SwitchHomeLoader = () => {
    const theme = useTheme();
    const isDark = theme.isDarkMode;

    const { riveFile } = useRiveFile(require('./assets/states-loading_home.riv'));
    const { riveViewRef, setHybridRef } = useRive();

    // `states-loading_home.riv` switches its `light_idle` / `dark_idle` states via
    // the `isDark` boolean INPUT on "State Machine 1"; the file has no data-binding
    // view model. Two things are required for the variant to switch on Android:
    //   1. Name the state machine explicitly (`stateMachineName` below) so the
    //      input is routed to the running machine — without it Android ignores the
    //      write and stays on the default variant.
    //   2. Only call `setBooleanInputValue` after `awaitViewReady()`. Do NOT call
    //      `play()` / `playIfNeeded()` / `reset()` here — those restart the machine
    //      from its Entry state and strand Android on the wrong (orange/default)
    //      frame.
    useEffect(() => {
        const view = riveViewRef;

        if (!view) {
            return;
        }

        let cancelled = false;
        void view.awaitViewReady().then(ready => {
            if (cancelled || !ready) {
                return;
            }
            view.setBooleanInputValue('isDark', isDark);
        });

        return () => {
            cancelled = true;
        };
    }, [isDark, riveViewRef]);

    return (
        <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
            {riveFile && (
                <RiveView
                    hybridRef={setHybridRef}
                    autoPlay
[...]

PR #343, opened 2026-07-16
fix(android): make play() start the configured state machine on cold start

mfazekas:
On Android, calling `play()` on a view mounted with `autoPlay={false}` didn't start the configured state machine: nothing is instanced at that point, so the wrapper fell through to rive-android's no-arg `controller.play()`, which starts the artboard's first raw animation plus its first state machine (ignoring `stateMachineName`). The result was wrong content (two timelines fighting) and, because the wrapper sampled the active state machine name before anything existed, subsequent input writes threw "View not configured. Could not find active state machine name". Reported in the follow-up of #332 ("play/playIfNeeded didn't work"); iOS was unaffected.

`play()` now starts the configured state machine explicitly on cold start (falling back to the artboard's first, matching what `autoplay` does) and records it as active. The no-arg resume path is kept for `pause()`/`play()` cycles and for animation-only files.

Includes a harness test that fails without the fix ("View not configured") and passes with it; verified on the Android emulator harness (full suite green) and on iOS (passes before and after).

Refs #332


HayesGordon (approved):
So annoying that the old Rive Android API is like this ☹️

Excited to get all of this out. Thanks for the fix!

PR #345, opened 2026-07-16
ci: file issue on publish failure and verify release is on npm

mfazekas:
The v0.4.15–v0.4.17 publish failures went unnoticed for three releases because releases are created via the release-please PAT, so GitHub's workflow-failure emails only go to the `rive-engineering` account. This adds two safeguards to the publish workflow:

1. After `npm publish`, poll the registry (up to 5 minutes) until the published version is actually visible, so a run only goes green once the package is really live.
2. If any step fails, automatically file an issue on the repo with a link to the failed run and cc the maintainers, so failures surface where people actually look.

Needs `issues: write` on the job, otherwise uses the default `GITHUB_TOKEN` — no new secrets.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,6 +10,7 @@ jobs:
     permissions:
       contents: read
       id-token: write
+      issues: write
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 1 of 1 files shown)


HayesGordon (approved):
LGTM!

PR #351, opened 2026-07-20
docs: adopt Rive's "new runtime" / "legacy runtime" terminology

mfazekas:
Aligns our wording with Rive's official docs, which call the CommandQueue-based runtimes the "New Runtime" (Apple) / "New Compose API" (Android) and the previous ones "legacy" — "experimental" was our own term and is now off-message.

1. Docs and spec JSDoc use the new/legacy vocabulary; user-facing error messages now say e.g. "not supported by the new runtime — use data binding".
2. API docs describe plain runtime behavior instead of comparing backends: the backend switch is for internal behavior-comparison testing only, so `getBackend()` is documented `@internal` (its `'experimental' | 'legacy'` return value is unchanged), the `updateReferencedAssets` warning no longer suggests switching to the legacy backend, and backend talk lives only in the contributor guide (`docs/runtime-backends.md`).
3. Adds `docs/migrating-to-0.5.md`: migrate to the async APIs on 0.4.x (full deprecated→replacement table), try the 0.5 beta, 0.5 stable keeps runtime deprecation warnings, and removal of the deprecated APIs is planned for 0.6.
4. The legacy runtime itself carries no removal-version commitment — "a future release once the new runtime has fully proven out (not necessarily 0.6)" — so the fallback stays available through a potentially breaking 0.6.

No behavior change; the only code deltas are two warning-message strings.

### DIFF
--- a/docs/migrating-to-0.5.md
+++ b/docs/migrating-to-0.5.md
@@ -46,5 +46,5 @@ deprecated calls keep working with runtime warnings.
 
 ## 4. 0.6
 
-The deprecated APIs and the legacy runtime are removed. Apps that completed
+Removal of the deprecated APIs is planned for 0.6. Apps that completed
 step 1 need no further changes.
--- a/docs/migrating-to-0.5.md
+++ b/docs/migrating-to-0.5.md
@@ -1,0 +1,50 @@
+# Migrating to 0.5 (the new Rive runtime)
+
+0.5 switches the library to Rive's new CommandQueue-based runtime
+([New Runtime](https://rive.app/docs/runtimes/apple/apple) on iOS,
+[New Compose API](https://rive.app/docs/runtimes/android) on Android) and
+completes the move to the async JS API. The recommended path:
+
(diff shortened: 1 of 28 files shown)


docs/migrating-to-0.5.md
```diff
@@ -0,0 +1,50 @@
...
+  handles; a typo'd path surfaces via the `getValueAsync()` rejection or the
+  `useRive*` hooks' `error` result instead of an `undefined` return.
+- `updateReferencedAssets` (runtime asset swapping) is not supported.
+
+## 3. 0.5 stable
+
+Same behavior as the beta: fully migrated apps run warning-free; remaining
+deprecated calls keep working with runtime warnings.
+
+## 4. 0.6
```
HayesGordon, line 47: For this one, we might need to say that 0.6 will potentially remove it completely. Or that some future version will. We might be forced to make some breaking change to 0.6, but at that time don't want to pull the trigger to remove the legacy fallback

HayesGordon (approved):
LGTM!

PR #294, opened 2026-06-15
feat: type-safe .riv schema system

mfazekas:
## Overview

Codegen pipeline that extracts artboard/state machine/ViewModel schemas from `.riv` files and surfaces them as TypeScript phantom types — no runtime overhead.

## How it works

**Step 1 — generate `.riv.d.ts` alongside each asset (committed to git):**
```sh
yarn rive-gen-types example/assets/rive/rewards.riv
yarn rive-gen-types --all example/assets/rive/  # batch
```

Uses `@rive-app/canvas` WASM to introspect the file and emits a declaration:
```ts
declare const asset: RiveAsset<{
  artboards: 'Main' | 'Lives 2';
  defaultArtboard: 'Main';
  stateMachines: { 'Main': 'State Machine 1' };
  viewModels: {
    'Rewards': {
      'Price_Value': 'number';
      'favourite_pet': 'enum:chipmunk|rat|frog|owl|cat|dog';
      'Coin': 'viewModel:Item_Icon_Value';
    };
  };
}>;
export default asset;
```

**Step 2 — use the typed asset:**

```tsx
import rewardsRiv from './assets/rive/rewards.riv';

// TypedRiveFile<T> — accepts RiveAsset directly, schema flows through
const { riveFile } = useRiveFile(rewardsRiv);

// Artboard/SM names constrained to schema
<RiveView file={riveFile} artboardName="Main" stateMachineName="State Machine 1" />

// viewModelName constrained to schema VMs
const { instance } = useViewModelInstance(riveFile, { viewModelName: 'Rewards' });

// Property paths constrained to the correct kind
useRiveNumber('Price_Value', instance);          // ✓
useRiveNumber('Coin', instance);                 // ✗ — Coin is a viewModel ref, not number
useRiveNumber('Coin/Item_Value', instance);      // ✓ — nested path
useRiveTrigger('Coin/Item_Value', instance);     // ✗ — Item_Value is number, not trigger

// Enum values constrained to the declared set
const { value } = useRiveEnum('favourite_pet', instance);
//    value: 'chipmunk' | 'rat' | 'frog' | 'owl' | 'cat' | 'dog'

// Inline types — no intermediate type aliases needed
function MyComponent({ file }: { file: TypedRiveFile<typeof rewardsRiv> }) { ... }
type RewardsInstance = TypedViewModelOf<typeof rewardsRiv, 'Rewards'>;
```

## What's included

[...]

PR #352, opened 2026-07-22
feat: type-safe .riv schema system

mfazekas:
Supersedes #294 (same feature, re-based onto main after the experimental-runtime merge in #134; squashed there since every commit conflicted with the runtime rewrite).

Codegen pipeline that extracts artboard/state machine/ViewModel schemas from `.riv` files and surfaces them as TypeScript phantom types — no runtime overhead. `yarn rive-gen-types example/assets/rive/rewards.riv` (or `--all`) emits a `.riv.d.ts` next to each asset; importing the asset then constrains `artboardName`/`stateMachineName`, `viewModelName`, property paths per kind, and enum value unions across `RiveView`, `useRiveFile`, `useViewModelInstance`, and the `useRive*` property hooks. See #294 for the full API walkthrough.

Changes relative to #294:

1. Merged with the experimental-runtime changes: `RiveFileFactory` keeps `getBackend()`, the `loadCdn` deprecation warning and optional `loadCdn?` signatures, now with the typed generics on top; `RiveView` keeps the new docs/`frameRate` while its props are generic over the schema.
2. Bumped `@rive-app/canvas` to 2.39.0 and regenerated all schemas — output is byte-identical, so the extractor works with the new runtime as is.
3. Generated schemas for assets added on main since the branch (`tabtest.riv`, `arbtboards-models-instances.riv`).

Note: rive-wasm 2.39.0 now exposes `ViewModelProperty.enumName`, so the "enum names not accessible" limitation from #294 is gone — a follow-up can emit named enum types instead of inlined value unions.


scripts/rive-gen-types.ts
```diff
@@ -0,0 +1,358 @@
...
+
+/** Per-file guard: a stalled WASM load() must fail loudly, never hang the batch. */
+const LOAD_TIMEOUT_MS = 30_000;
+
+async function extractSchema(input: string): Promise<Schema> {
+  const bytes =
+    input.startsWith('http://') || input.startsWith('https://')
+      ? new Uint8Array(await (await fetch(input)).arrayBuffer())
+      : new Uint8Array(readFileSync(input));
+
```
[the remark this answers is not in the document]
mfazekas: Addressed in 5770418 — URL fetches now check res.ok and fail early with `HTTP <status>: <url>`.

scripts/rive-gen-types.ts
```diff
@@ -0,0 +1,358 @@
...
+ *
+ * Usage:
+ *   bun scripts/rive-gen-types.ts <path-or-url>          # writes <file>.riv.d.ts next to the source
+ *   bun scripts/rive-gen-types.ts <path> --out <out.ts>  # write a standalone schema .ts instead
+ *   bun scripts/rive-gen-types.ts --all <directory>      # generate for every .riv in a directory
+ *
+ * After generation, TypeScript resolves the .riv.d.ts automatically:
+ *   import gameRiv from './assets/game.riv';              // typed as RiveAsset<GameSchema>
+ *   const file = await RiveFileFactory.fromSource(gameRiv); // TypedRiveFile<GameSchema> — T inferred
+ */
```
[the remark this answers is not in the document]
mfazekas: Addressed in 5770418 — the header and usage text now state that URL inputs require `--out`, and the `fromSource` example passes the required `referencedAssets` argument (`undefined`).

scripts/rive-gen-types.ts
```diff
@@ -0,0 +1,358 @@
...
+    // Standalone mode: generate a named schema type, not a .d.ts
+    const outPath = resolve(process.cwd(), args[outIdx + 1]!);
+    const baseName = basename(input, '.riv').replace(/[^a-zA-Z0-9]/g, '_');
+    const nameIdx = args.indexOf('--name');
+    const typeName =
+      nameIdx !== -1
+        ? args[nameIdx + 1]!
+        : baseName.charAt(0).toUpperCase() + baseName.slice(1) + 'Schema';
+    await generate(input, outPath, 'standalone', typeName);
+  } else {
```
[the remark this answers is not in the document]
mfazekas: Addressed in 5770418 — `--out` and `--name` now validate that a value follows and print a clear one-line error instead of crashing.

.github/actions/setup/action.yml
```diff
@@ -9,6 +9,9 @@ runs:
       with:
         node-version-file: .nvmrc
 
+    - name: Setup Bun
```
HayesGordon, line 12: Not against it, but just want to make sure we're intentional, as all of the current workflows locally require `yarn`.

scripts/rive-extract-schema.ts
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env bun
```
HayesGordon: Same here. Will we now force users to use Bun? Not sure if there is a clean workaround to support more (or if Bun and NPM can be used interchangeably).

HayesGordon (dismissed):
This is amazing!! A lot of people will be really excited for this. A feature request is to also generate asset names (images, fonts, audio) for out-of-band handling. Some questions around using Bun, and wanting to understand the decision around it in case it comes up as a question/restraint.

.github/actions/setup/action.yml
mfazekas: It's kind of historic, we have node 20 in nvmrc it cannot run .ts directly. And for development running .ts directly  without build step was essential. But newer Node is a real option 22.23.1  or 24.15.0+.  So bun is not really needed - will remove from the PR.

scripts/rive-extract-schema.ts
mfazekas: Good catch, we don't really need shebang. We'll not invoke scripts/rive-extract-schema.ts directly but via npx or bunx whatever.

HayesGordon (approved):
LGTM! This is great!

PR #359, opened 2026-07-29
feat: generate referenced asset names in .riv schemas, type referencedAssets

mfazekas:
Stacked on #385 (named enums): main ← #385 ← this. Implements the review feature request: generate asset names (images, fonts, audio) for out-of-band handling.

Generated schemas gain a `referencedAssets` record — the asset's runtime `uniqueName` (the documented `referencedAssets` key form, e.g. `'Inter-594377'`) → `'image' | 'font' | 'audio'`. Only non-embedded assets are listed: referenced and CDN-hosted ones are exactly what an app must supply on the new runtime, which doesn't fetch CDN assets (embedded assets need nothing, so they're omitted). Extraction happens in the existing WASM `load()` asset-loader callback — the only place asset metadata is visible — so no extra pass.

On the typing side, `useRiveFile` and `RiveFileFactory.fromSource` constrain `referencedAssets` for schema-typed inputs:

1. keys must name an asset in the file,
2. each entry's declared `type` must match the asset's actual kind (mislabeling a font as `image` is a compile error — useful since the new runtime can't infer types and relies on this hint),
3. `RiveImage` objects are only accepted for image assets,
4. files with no referenced assets reject all keys.

`useRiveFile`'s and `fromSource`'s typed/untyped overload pairs are each merged into one generic signature so an invalid key is a hard error instead of silently falling through to the untyped number-input overload (the same fallthrough class fixed in #355 for `viewModelName` — caught here by sabotage-testing the example before writing the types). Unknown schemas and plain inputs degrade to the untyped `ReferencedAssets` exactly as before.

The out-of-band example now uses the typed asset import as a live showcase; `out_of_band.riv.d.ts` shows the emitted shape (3 referenced + 3 CDN assets). All 31 committed schemas regenerated. Coverage: tsd pins for every mistake class plus backward-compat, bun unit tests for `classifyAsset`/emit, and the emit fuzzer extended with hostile asset names.

Runtime behavior is unchanged (pure codegen + types); semantics target the new runtime per the request — legacy is out of scope.

Verified compiler errors per mistake class (captured from a real probe against the repo tsconfig):

1. Unknown key: `TS2353: Object literal may only specify known properties, and ''Inter-59437'' does not exist in type '{ "Inter-594377"?: …'` — the valid keys are listed in the error and autocomplete in the editor.
2. Wrong asset type: `Type '"image"' is not assignable to type '"font"'`.
3. `RiveImage` for a non-image asset: `Type 'RiveImage' is not assignable to type 'ReferencedAssetSource & { type?: "font" }'`.
4. Entry for a file with no referenced assets: `Type '{ source: number; type: string; }' is not assignable to type '"this .riv file has no referenced (out-of-band) assets"'` — the rejection type is a literal message so the error explains itself (a plain `never` gave an unexplained "not assignable to never").

Backward compat is pinned by tsd for every untyped input form — url string, `{ uri }`, `require()` (`any`), and plain `number` — all keep accepting arbitrary asset keys.

Schemas generated before this change have no `referencedAssets` key and no longer satisfy `RiveFileSchema` (a compile error, not a silent downgrade) — regenerate with `npx rive-gen-types`.

Image view model properties are now `'assetImage'` (RML's `ViewModelPropertyAssetImage`, our native `ViewModelPropertyType`) instead of the web runtime's `'image'`, which collided with the `'image'` asset kind this PR introduces.


HayesGordon (approved):
LGTM!

PR #364, opened 2026-08-16
feat!: remove onStop callback in favor of data-binding triggers

mfazekas:
Removes the `onStop` prop introduced in #353 (shipped only in `v0.5.0-beta.1`). Per the discussion with the Rive team, state-machine settling is intentionally internal — rive-android 11.8.0 already removed the public `settledFlow` this was built on — so app behavior shouldn't depend on it.

The replacement is a data-binding trigger fired by the state machine when the timeline completes (see [this community file](https://rive.app/community/files/28526-53929-fire-a-trigger-when-timeline-finishes) for the editor setup). Shipped here as a README section (including the exit-time-100% gotcha), a `finished_trigger.riv` asset with a "Finished Trigger" demo page in the example app, and a harness test asserting the trigger fires exactly once and only after the timeline actually plays — on all four backends, legacy included.

The Android render loop keeps the internal stop-advancing-when-settled optimization from #353; the iOS settled machinery is removed entirely (the SDK's own view already stops advancing at rest).


mfazekas:
Responses to the Copilot review comments (posting here — inline replies aren't available to me):

**README snippet imports** — fixed in 92b19bf. While adding them I also caught that the snippet called `useViewModelInstance(riveFile)` with no options, which resolves to the deprecated overload ("creates the instance synchronously via deprecated runtime APIs that block the JS thread"); it now passes `{ async: true }`, matching the demo page.

**`!settled` guard in `observeSettled()`** — leaving as is. The "emits repeatedly while at rest" premise no longer holds after this PR: the loop stops calling `advanceStateMachine` once `settled` is true, and the command server only emits on an advance whose `advanceAndApply` returns false — so emissions stop too, apart from the one or two already in flight before the collector catches up. Guarding a couple of idempotent volatile writes of a constant isn't worth the extra condition, which mostly makes a reader wonder whether the write is non-idempotent.

Worth noting the guard doesn't buy the safety it looks like it does: if an emission queued before a pointer event is delivered after `settled` was reset to `false`, the `!settled` check passes and it sets `settled = true` anyway, parking the view until the next interaction. That's a real (rare, self-healing) race on the pinned 11.7.2, and it's fixed upstream rather than here — the restored `settledFlow` is backed by a generational store that rejects callbacks from before the latest unsettled boundary. A follow-up PR bumps rive-android for exactly that.

HayesGordon (approved):
LGTM. I'd prefer for us to keep this, even if it's just for dev/testing purposes. But we can revisit. There is other work that we want to do for all the runtimes to move a way from the idea of play/pause and expose finer control to users

PR #372, opened 2026-08-21
test(ios): failing harness test for blank RiveView after Fabric recreates the view

mfazekas:
Adds a harness test for the blank-view bug reported in #365. Fabric deletes a component view when its subtree stops being mounted — `display: 'none'`, or a `react-native-screens` screen frozen by `enableFreeze(true)` — and recreates it from the same, unchanged ShadowNode. On iOS the recreated view is never configured, because nitro's `isDirty` prop flags live on the shared `Props` object and the first view instance already consumed them: no file, blank view forever, and `play()` on the ref JS still holds is a no-op.

The test fails on iOS today and passes on Android, which doesn't recreate the view. The fix is the nitro 0.37.x bump (mrousavy/nitro#1503 + #1506 + #1510 replace the `isDirty` scheme with an old/new props diff), landing in a follow-up commit here.

On the shape of the test: it toggles `display: 'none'` on the parent rather than pulling in `react-freeze`, since that produces the same delete/recreate with no extra dependency. The oracle needed some care — the harness has no screenshot API, and the stale ref still answers `getViewModelInstance()`, `awaitViewReady()` and `play()` quite happily on the broken build. What does discriminate is a trigger fired on the bound view model instance: it only reaches its listener while a live view advances the state machine. The same assertion runs before the hide/show as a control, so a failure means the view stopped working rather than the probe never having worked.

Verified by rebuilding and reinstalling the app for each arm: iOS without the fix fails, iOS with #365's patch applied passes, Android passes.

<details>
<summary>Manual reproducer page (drop into any new-arch app)</summary>

Both toggles produce the same Fabric delete/recreate: `react-freeze` is what `enableFreeze(true)` uses, and `display: 'none'` needs no dependency at all. Hit **Hide + Show** — on an unpatched build the box goes solid black and stays there, `hybridRef fires` never increments, and `play()` resolves without doing anything.

```tsx
import { useRef, useState } from 'react';
import { View, Text, StyleSheet, Pressable } from 'react-native';
// ships with react-native-screens
import { Freeze } from 'react-freeze';
import {
  RiveView,
  useRiveFile,
  Fit,
  type RiveViewRef,

type Mode = 'freeze' | 'display';

export default function FreezeRemountBlank() {
  const [mode, setMode] = useState<Mode>('freeze');
  const [hidden, setHidden] = useState(false);
  const [refFires, setRefFires] = useState(0);
  const [playResult, setPlayResult] = useState('—');
  const viewRef = useRef<RiveViewRef | null>(null);
  const { riveFile, error } = useRiveFile(require('./rewards.riv'));

  const rive = riveFile && (
    <RiveView
      file={riveFile}
      autoPlay={true}
      fit={Fit.Cover}
      style={styles.rive}
      hybridRef={{
        f: (ref) => {
          viewRef.current = ref;
          setRefFires((n) => n + 1);
        },
      }}
    />
  );

  return (
    <View style={styles.container}>
      <View style={styles.row}>
        {(['freeze', 'display'] as Mode[]).map((m) => (
          <Pressable
            key={m}
            onPress={() => {
              setHidden(false);
              setMode(m);
            }}
            style={[styles.button, mode === m && styles.buttonActive]}
          >
            <Text style={styles.buttonText}>
              {m === 'freeze' ? 'react-freeze' : 'display:none'}
            </Text>
          </Pressable>
        ))}
      </View>

      <View style={styles.row}>
        <Pressable onPress={() => setHidden((h) => !h)} style={styles.button}>
          <Text style={styles.buttonText}>{hidden ? 'Show' : 'Hide'}</Text>
        </Pressable>
        <Pressable
          onPress={() => {
            setHidden(true);
            setTimeout(() => setHidden(false), 300);
          }}
          style={styles.button}
        >
          <Text style={styles.buttonText}>Hide + Show</Text>
        </Pressable>
        <Pressable
          onPress={() => {
            setPlayResult('pending');
            viewRef.current
              ?.play()
              .then(() => setPlayResult('resolved'))
              .catch((e: unknown) => setPlayResult(`rejected: ${String(e)}`));
          }}
          style={styles.button}
        >
          <Text style={styles.buttonText}>play()</Text>
        </Pressable>
      </View>

      <Text style={styles.status}>
        hidden: {String(hidden)} · hybridRef fires: {refFires} · play():{' '}
        {playResult}
      </Text>
      {error != null && (
        <Text style={styles.status}>Error: {String(error)}</Text>
      )}

      <View style={styles.stage}>
        {mode === 'freeze' ? (
          <Freeze freeze={hidden}>{rive}</Freeze>
        ) : (
          <View style={hidden ? styles.hiddenBox : styles.box}>{rive}</View>
        )}
      </View>
    </View>
  );
}

const styles = StyleSheet.create({
  container: { flex: 1, backgroundColor: '#fff', padding: 12 },
  row: { flexDirection: 'row', gap: 8, marginBottom: 8 },
  button: {
    paddingHorizontal: 12,
    paddingVertical: 8,
    borderRadius: 6,
    backgroundColor: '#444',
  },
  buttonActive: { backgroundColor: '#0a7' },
  buttonText: { color: '#fff', fontWeight: '600' },
  status: { marginBottom: 8, color: '#333' },
  stage: { width: 260, height: 260, backgroundColor: '#000' },
  box: { flex: 1 },
  hiddenBox: { flex: 1, display: 'none' },
  rive: { flex: 1 },
});
```

</details>

## Nitro 0.37.1

The bump is the fix. Generated `updateProps` now diffs the new Props snapshot against the old one instead of consuming shared `isDirty` flags, so a recreated view re-applies every provided prop. Full iOS harness against a rebuilt app: 30 suites, 210 tests green, `view-recreate` among them. The peer range moves to `>=0.37.0 <0.38`, so consumers have to move too. (Examples pin 0.37.1, which adds mrousavy/nitro#1521 and ArrayBuffer memory-pressure reporting; the generated output is identical.)

Three things ride along with it:

1. Prop parsing moved into Nitro core, taking away the parse site the `nitrogen-postprocess.ts` shim for mrousavy/nitro#1184 used to patch — and that bug is *not* fixed in 0.37.x (mrousavy/nitro#1515 is still open): clearing an optional prop still throws mid-commit (`RiveView.layoutScaleFactor: Value is null, expected a number`, and `RiveView.dataBind: Cannot convert "null" to any type in variant<...>`). The shim is reinstated one level up: postprocess now injects a `parseNullAsCleared` wrapper into the generated Props constructor, which reads the raw value through the public `RawPropsCompat::at` and parses a null-valued optional prop as a cleared one before nitro ever sees it. The fix still ships inside the package and works on stock nitro, so consumers don't need to patch nitro. `optional-prop-clear.harness.tsx` covers both props and fails without the injection.
2. The generated `EventPropertiesOutput` variant is now ordered `boolean | number | string`, so the two hand-written helpers swap `.second` and `.third`.
3. glog. Nitro exposes React's renderer headers in the `NitroModules` modulemap, and they reach `<glog/logging.h>`, which includes headers from inside `namespace google` — illegal once glog is a module, so every target that builds the NitroModules module fails to compile. Same class of breakage as mrousavy/nitro#1520; 0.37.1 (mrousavy/nitro#1521) fixed only the `cxxreact` half — `ReactProp.hpp` still reaches glog through `RawValue.h`, verified: stock 0.37.1 fails identically. Nothing in React Native `@import`s glog, so the workaround drops glog's modulemap and includes it textually. It ships as `ios/rive_textual_glog.rb` and installs itself: `RNRive.podspec` requires it and it hooks the CocoaPods installer (a dependency's podspec gets no post-install hook of its own, but podspecs are plain Ruby evaluated inside every `pod install`), so consumer apps build with no Podfile change — verified by removing the example Podfile's hook and diffing the generated xcconfig tree, byte-identical. Comes out once Nitro keeps the renderer headers out of its public modulemap.

### DIFF
--- a/example/__tests__/view-recreate.harness.tsx
+++ b/example/__tests__/view-recreate.harness.tsx
@@ -1,0 +1,145 @@
+import {
+  describe,
+  it,
+  expect,
+  render,
+  waitFor,
+  cleanup,
+} from 'react-native-harness';
+import { useState } from 'react';
+import { View } from 'react-native';
+import {
+  RiveView,
+  RiveFileFactory,
+  Fit,
+  type RiveFile,
+  type RiveViewRef,
+} from '@rive-app/react-native';
+import type { ViewModelInstance } from '@rive-app/react-native';
+
+/**
+ * Fabric deletes a component view when its subtree stops being mounted
+ * (`display: 'none'`, or a react-native-screens screen frozen by
+ * `enableFreeze(true)`) and recreates it from the same, unchanged ShadowNode
+ * when the subtree comes back. The recreated view has to be configured from
+ * those props.
+ *
+ * On iOS it was not: nitro tracks props with `isDirty` flags stored on the
+ * shared Props object and clears them once applied, so the second view
+ * instance was handed a props object whose flags the first instance had
+ * already consumed. It never received its file, stayed blank forever, and the
+ * ref JS holds kept pointing at the dead view. See PR #365.
+ */
+
+const QUICK_START = require('../assets/rive/quick_start.riv');
+
+function expectDefined<T>(value: T): asserts value is NonNullable<T> {
+  expect(value).toBeDefined();
+}
+
+type TestContext = {
+  ref: RiveViewRef | null;
+  error: string | null;
+  setHidden: ((hidden: boolean) => void) | null;
+};
+
+// The visibility state lives here so that flipping it re-renders this
+// component only, leaving the RiveView's own ShadowNode untouched — the
+// situation a frozen screen creates.
+function HideableRive({
+  context,
+  file,
+  instance,
+}: {
+  context: TestContext;
+  file: RiveFile;
+  instance: ViewModelInstance;
+}) {
+  const [hidden, setHidden] = useState(false);
+  context.setHidden = setHidden;
+
+  return (
+    <View
+      style={{ width: 200, height: 200, display: hidden ? 'none' : 'flex' }}
+    >
+      <RiveView
+        hybridRef={{
+          f: (ref: RiveViewRef | null) => {
+            context.ref = ref;
+          },
+        }}
+        style={{ flex: 1 }}
+        file={file}
+        autoPlay={true}
+        dataBind={instance}
+        fit={Fit.Contain}
+        stateMachineName="State Machine 1"
+        onError={(e) => {
+          context.error = e.message;
+        }}
+      />
+    </View>
+  );
+}
+
+// A trigger only reaches its listener while a live view advances the state
+// machine the instance is bound to, which is what makes this a usable answer
+// to "is the view still driving this data binding?".
+async function triggerReachesListener(
+  instance: ViewModelInstance
+): Promise<boolean> {
+  const trigger = instance.triggerProperty('gameOver');
+  expectDefined(trigger);
+  let fired = false;
+  const removeListener = trigger.addListener(() => {
+    fired = true;
+  });
+  trigger.trigger();
+  await waitFor(
+    () => {
+      expect(fired).toBe(true);
+    },
+    { timeout: 1000 }
+  ).catch(() => {});
+  removeListener();
+  trigger.dispose();
+  return fired;
+}
+
+describe('view recreated by Fabric (PR #365)', () => {
+  it('keeps driving its data binding after hide/show', async () => {
+    const file = await RiveFileFactory.fromSource(QUICK_START, undefined);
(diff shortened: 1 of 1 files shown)


rlods:
@mfazekas do you have any update about this fix? 👋 
and do you know when next release will be submitted?

mfazekas:
> @mfazekas do you have any update about this fix? 👋 and do you know when next release will be submitted?

#375 was merged, and will cut a new beta release soon

PR #375, opened 2026-09-01
fix: force-apply props when Fabric recreates a RiveView

mfazekas:
Fabric can delete a component view and recreate it from the same, unchanged ShadowNode — `react-freeze`d screens (what `enableFreeze(true)` uses) or a plain `display: 'none'` toggle both trigger it. nitro 0.35's per-prop `isDirty` flags live on the shared Props object and were already consumed by the previous view instance, so the recreated `RiveView` stays blank: no file, no artboard, and the ref JS holds points at a dead view.

Closes #365 (no standalone issue exists for this bug — the report lives in that PR). This applies #365's fix through `nitrogen-postprocess.ts` so `yarn nitrogen` doesn't wipe it: generated `updateProps` force-applies every prop the first time a view instance updates. The new `view-recreate.harness.tsx` fails on unpatched main (a `gameOver` trigger fired after the hide/show cycle never reaches its listener) and passes with the patch; full iOS harness is green with it (29 suites, 209 tests). Android is unaffected — it never recreates the view — and its generated code is untouched.

nitro 0.37 fixes this properly by diffing old vs new props, but the bump (#372) is parked until the nitro-side issues it surfaced (glog modulemap break, mrousavy/nitro#1184) are fixed upstream. This shim and its injector are deleted with that bump.


HayesGordon (approved):
LGTM!

PR #380, opened 2026-09-09
fix: bump nitro to 0.36.5 so the library works on React Native 0.87

### LINKED ISSUE #378: RN 0.87 incompatible: nitro peer capped at <0.36, but the installJSIBindingsWithRuntime:callInvoker: fix ships in nitro 0.36.2+
### Description

`@rive-app/react-native` caps `react-native-nitro-modules` at `>=0.35.10 <0.36`, but React Native 0.87 requires nitro **0.36.2+**. On RN 0.87 the app dies at startup before any JS runs:

```
[runtime not ready]: Error: Failed to install Nitro: installJSIBindingsWithRuntime: was not called - JSI Bindings could not be installed!
[runtime not ready]: Invariant Violation: "arthosMobile" has not been registered. This can happen if:
* Metro (the local dev server) is run from the wrong folder. Check if Metro is running, stop it and restart it in the current project.
* A module failed to load due to an error and `AppRegistry.registerComponent` wasn't called.
```

**Root cause (not Rive's code — the version cap):**

RN 0.87 changed the iOS TurboModule JSI hook. `RCTTurboModuleManager.mm` now only calls the two-argument selector, guarded by `respondsToSelector:`, with no fallback to the old one-argument form:

```objc
// react-native/ReactCommon/react/nativemodule/core/platform/ios/ReactCommon/RCTTurboModuleManager.mm:351
if ([module respondsToSelector:@selector(installJSIBindingsWithRuntime:callInvoker:)]) {
  [(id<RCTTurboModuleWithJSIBindings>)module installJSIBindingsWithRuntime:*runtime callInvoker:_jsInvoker];
}
```

`react-native-nitro-modules` 0.35.x implements only `installJSIBindingsWithRuntime:` (one arg), so it is never invoked, `global.NitroModulesProxy` is never created, and every Nitro module — Rive included — fails at startup. This was fixed upstream in margelo/nitro#1435 / #1437, first released in **nitro 0.36.2** (0.36.1 does not have it; verified in the published tarballs).

Because `@rive-app/react-native` pins `<0.36`, a project that needs RN 0.87 cannot install a working nitro version without going against Rive's declared peer range.

**Ask:** bump `react-native-nitro-modules` / `nitrogen` to `>=0.36.2` (0.37.x is current), regenerate `nitrogen/generated/**`, and widen the peer range. Note the generated code is vendored in the published package, so consumers cannot fix this on their own side by upgrading nitro alone — the shipped specs are what get compiled.

**Additional data point:** I forced nitro `0.37.1` locally (peer range ignored) with `@rive-app/react-native@0.4.20`. It **builds and runs correctly on iOS** — Nitro installs, all Rive HybridObjects (`RiveView`, `RiveFile`, `RiveFileFactory`, `RiveRuntime`, `RiveFontConfig`, `RiveImageFactory`) register, and the Rive animation renders. The only compiler output is a deprecation wall from the 0.35-era generated specs:

```
@rive-app/react-native/nitrogen/generated/shared/c++/views/HybridRiveViewComponent.hpp:53:5:
warning: 'CachedProp<std::optional<std::string>>' is deprecated: CachedProp<T> is deprecated.
Update nitrogen and re-generate your specs to use ReactProp<T> instead.
```

So the cap looks like a policy/version-guard issue rather than a genuine incompatibility — though the `CachedProp` → `ReactProp` migration is presumably why regeneration is wanted before widening it.

### Previous working version

No version works on RN 0.87. `0.4.20` (and every earlier release) works fine on RN 0.86 with nitro 0.35.10.

### Reproduction steps / code

No `.riv`-specific repro is needed — this is a startup-time JSI/TurboModule failure that happens before any Rive API is called.

1. RN 0.87.0 app, New Architecture enabled, Hermes, bare (non-Expo).
2. `npm i @rive-app/react-native@0.4.20` — resolves `react-native-nitro-modules@0.35.10` (the newest version its peer range allows).
3. `cd ios && pod install && npx react-native run-ios`.
4. Native build succeeds; the app red-screens on launch with the two errors above.
5. Set `react-native-nitro-modules` to `0.37.1` (ignoring the peer range), `pod install`, rebuild → the app starts and Rive works.

### `@rive-app/react-native` version

0.4.20 (also reproduced on 0.4.11; `main` @ 0.5.0-beta.3 has the same `>=0.35.10 <0.36` peer range)

### Platform

iOS only

### React Native version

0.87.0

### React Native environment info

```shell
System:
  OS: macOS 26.6.2
  CPU: (14) arm64 Apple M4 Pro
Binaries:
  Node: 22.22.2
  npm: 10.9.7
  Watchman: 2026.07.27.00
Managers:
  CocoaPods: 1.16.2
IDEs:
  Xcode: 26.6/17F113
Languages:
  Java: 17.0.20.1
  Ruby: 3.3.0
npmPackages:
  "@react-native-community/cli": 20.2.0
  react: 19.2.8
  react-native: 0.87.0
iOS:
  hermesEnabled: true
  newArchEnabled: true
```

### Expo setup

No — bare React Native

### iOS device & OS version

iOS Simulator — iPhone 17 Pro, iOS 26.x (Xcode 26.6)

### Additional context

- Upstream nitro fix: https://github.com/margelo/nitro/issues/1435 (closed by margelo/nitro#1437, shipped in `react-native-nitro-modules@0.36.2`).
- Affected consumer PR: https://github.com/dcodeit/arthos-mobile/pull/543 — our RN 0.87 upgrade is blocked on this.
- Related, but separate, RN 0.87 / AGP 9 Android problem in this repo: #377.

mfazekas:
Fixes #378.

React Native 0.87 only calls the two-argument `installJSIBindingsWithRuntime:callInvoker:` TurboModule hook, which nitro added in 0.36.2. With the previous `<0.36` peer range no Nitro module could install and the app never registered. Reproduced in a bare RN 0.87.0 app with the published 0.5.0-beta.3: native build succeeds, blank screen, `Failed to install Nitro: installJSIBindingsWithRuntime: was not called` in the log.

This moves nitrogen and `react-native-nitro-modules` to 0.36.5, regenerates the specs and widens the peer range to `>=0.35.10 <0.37`. The floor is unchanged because the regenerated specs still build and run against nitro 0.35.10 (verified on both platforms plus three harness suites), so existing consumers are not forced to bump; RN 0.87 users need 0.36.2 or newer, which the README now says. The cap stays below 0.37 on purpose: nitro 0.37 moves prop parsing into its core, where clearing an optional prop still throws (mrousavy/nitro#1515 is unmerged), and our existing null-prop shim only anchors on the 0.36 generated code.

nitrogen 0.36 changes two things the generated view code relies on. Both are handled in `scripts/nitrogen-postprocess.ts`, the same way nitro 0.37 fixes them upstream:

1. The component descriptor is built with a plain `RawPropsParser()`, which only keeps props as `jsi::Value`s on RN >= 0.85. Older versions convert every prop to `folly::dynamic` and the generated Props constructor aborts on the first RiveView mount. The postprocess forces the JSI parser where the flag still exists.
2. The component view never sets `_props`, which RN 0.87 turns into a fatal `RCTViewComponentView` assertion on debug builds. The postprocess initializes it to the component's default props in `init`.

The `EventPropertiesOutput` variant is now ordered `boolean | number | string`, so the hand-written Swift and Kotlin helpers swap their cases.

Verified: example iOS (RN 0.80.3) harness 28 suites / 208 tests, Android harness 28 suites / 208 tests, and a bare RN 0.87.0 app with the packed tarball + nitro 0.36.5 installs Nitro and renders `rewards.riv`. The same bump should be backported to release-v0.4 so npm `latest` works on RN 0.87 too.


HayesGordon (approved):
LGTM!

PR #381, opened 2026-09-10
feat: expose GPU Canvas (3D) on iOS and Android

mfazekas:
Exposes GPU Canvas (required for 3D content) on the new runtime as `RiveRuntime.setGPUCanvasEnabled(true)` / `isGPUCanvasEnabled()`, and bumps rive-ios to 6.26.0 and rive-android to 11.12.1, the first releases that support it. The setting is process-wide and must be set before the first Rive file loads; a later call logs a warning and is ignored. No-op on the legacy runtime.

iOS uses `Worker(configuration: .init(enableGPUCanvas:))`. Android uses rive-android's experimental deferred renderer, whose `CommandQueue.createDeferred` is Kotlin-internal, so `DeferredRiveWorker.kt` resolves it by reflection; if the rive-android in use has no such entry point the worker is created without GPU Canvas and a warning is logged.

The 11.12.x bump also picks up the fix for rive-app/rive-android#470 (debug-build abort when reading a nonexistent boolean property, found while testing 11.12.0); the harness test now covers nested misses too.

Example page `demos/GPUCanvasDemo.tsx` renders `ore.riv` (rive-android's sample, MIT). Verified on both platforms: harness green (iOS 29/29, Android 29/29), the 3D scene renders with the flag and is black without it. Docs in `docs/runtime-backends.md`.


HayesGordon (approved):
LGTM!

PR #383, opened 2026-09-21
fix(android): fall back with a warning when GPU Canvas is unavailable

mfazekas:
Stacked on #381. The GPU Canvas worker on Android was reached through a Java shim that named rive-android's Kotlin-internal `createDeferred$kotlin_release` at compile time, so any rive-android without that exact symbol (an older `Rive_RiveRuntimeAndroidVersion` override, or an upstream rename) failed the build.

`DeferredRiveWorker` now looks the method up by reflection, accepting the mangled name or a plain `createDeferred` (so it keeps working if upstream makes it public). When it is missing, the shared worker is created without GPU Canvas, a warning goes through `RiveLog`, and `RiveRuntime.isGPUCanvasEnabled()` reports `false` from then on. A real worker-init failure still throws.

Verified on the Pixel 6 emulator with the `ore.riv` demo: on 11.12.1 the 3D scene renders as before; on 11.11.0 (no deferred worker, previously a compile error) the app builds, logs the warning and runs without GPU Canvas.


mfazekas:
Folded into #381 (its branch now includes this commit) so the feature is reviewed once with the fallback in place.

